PR #39, opened 2026-02-27
Auth: backend and frontend admin authentication

Thoosje:
## feat: JWT authentication

Implements email/password authentication with JWT tokens.

### Changes

**Database**
- Added `users` migration with `email`, `password_hash`, `username`, `created_at` columns
- Added `sessions` migration with `refresh_token`
- Added `by_email` query to `UserRepo`
- Updated `User` model to include `email` and `password_hash`

**API**
- `POST /auth/login` — validates email/password, returns signed JWT
- `POST /admin/me` — protected route, returns admin info
- `POST /auth/refresh` — creates new `access_token` from `refresh_token`
- `POST /auth/logout` — removes `refresh_token` from DB and deletes sessions
- Password verification using `argon2`
- JWT signing using `jsonwebtoken` with `rust_crypto` backend
- `JWT_SECRET` loaded from environment variable

**Auth middleware**
- `AuthUser` extractor — validates `access_token` cookie token on any protected route
- Added `AppError::Unauthorized` variant mapping to 401

**Fontend**
- Edited `proxy.ts` to check for authenticated routes and check if the token exists
- Added axios api wrapper
- Added Tanstack Query `useAuth.ts` wrapper to cache user data
- Created `LoginForm.tsx` component

---

## TODO: Wire up in Next.js frontend

- [x] Add `POST /login` call on login form submit
- [x] Store returned JWT in cookie (prefer `httpOnly` cookie for security)
- [x]  Create an `axios`/`fetch` wrapper that attaches `Authorization: Bearer <token>` header to all API requests
- [x] Handle 401 responses globally — redirect to `/login`
- [x] Protect frontend routes using Next.js middleware (`proxy.ts`) that checks for a valid token
- [x] Add a logout button that clears the token
- [x] Check if token is still valid in frontend by checking age, and possibly even add hartbeat endpoint with redis to make it possible to blacklist tokens when logging out (and later reseting password)
- [x] Instead of returning JWT in json from backend return as cookie to prevent XXS attacks


backend/migrations/20260227093503_add_auth.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE users ( -- Call this users or admin_users?
+   id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
hanneskt: suggestion: use uuidv7 instead

We've been using UUIDv7 in other tables because it's time-sortable. While it's not really necessary here, it seems like a good idea to keep it consistent.

The function in postgres is `uuidv7()`

backend/migrations/20260227093503_auth.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE users ( -- Call this users or admin_users?
+   id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+   username TEXT NOT NULL,
+   email TEXT NOT NULL UNIQUE,
```
hanneskt, line 4: question: are we going to have an email service? 

it seems weird to store an email if we aren't going to use it. we can't verify it, we can't send password reset mails, etc. It might create the expectation that that is possible if we ask for an email.

suggestion: It seems more logical to have the username be unique and use that to log in than an email

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub sub: String,
+    pub email: String,
+    pub exp: usize,
+}
+
+pub struct AuthHandler;
+
+impl AuthHandler {
+    pub async fn login(
+        State(state): State<AppState>,
```
hanneskt: suggestion: use `db: Database` instead of the state to access the database

backend/src/extractors/auth.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    async fn from_request_parts(parts: &mut Parts, state: &AppState) -> Result<Self, AppError> {
+        // Extract the Authorization header
+        let auth_header = parts
+            .headers
+            .get("Authorization")
+            .and_then(|v| v.to_str().ok())
+            .ok_or(AppError::Unauthorized)?;
+
+        // Strip "Bearer "
```
hanneskt: chore: remove these comments

backend/src/extractors/auth.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+impl FromRequestParts<AppState> for AuthUser {
+    type Rejection = AppError;
+
+    async fn from_request_parts(parts: &mut Parts, state: &AppState) -> Result<Self, AppError> {
+        // Extract the Authorization header
+        let auth_header = parts
+            .headers
+            .get("Authorization")
```
hanneskt: suggestion: use `AUTHORIZATION` (from `header::AUTHORIZATION`) instead of a string here

backend/src/extractors/auth.rs
```diff
@@ -0,0 +1,39 @@
...
+
+pub struct AuthUser {
+    pub user_id: String,
+    pub email: String,
+}
+
+impl FromRequestParts<AppState> for AuthUser {
+    type Rejection = AppError;
+
+    async fn from_request_parts(parts: &mut Parts, state: &AppState) -> Result<Self, AppError> {
```
hanneskt, line 16: suggestion: maybe we can use TypedHeaders instead of manually parsing the headers?

You can see an example here https://docs.rs/axum-extra/latest/axum_extra/struct.TypedHeader.html

You can use it here like `let header = TypedHeader::<Authorization<Bearer>>::from_request_parts(parts, state).await;`

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,75 @@
...
+        Json(payload): Json<LoginRequest>,
+    ) -> Result<Json<AuthResponse>, AppError> {
+        tracing::info!("email: {}", payload.email);
+        tracing::info!("password: {}", payload.password);
```
hanneskt: issue: never log passwords

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+pub struct AuthHandler;
+
+impl AuthHandler {
+    pub async fn login(
+        State(state): State<AppState>,
+        Json(payload): Json<LoginRequest>,
+    ) -> Result<Json<AuthResponse>, AppError> {
+        tracing::info!("email: {}", payload.email);
```
hanneskt: suggestion: use the `debug!` log level here

this is fine, but logging of values belongs more in the debug log level

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,75 @@
...
+        tracing::info!("password: {}", payload.password);
+
+        // 1. Fetch user by email
+        let user = state.db.users()
+            .by_email(&payload.email)
+            .await
+            .map_err(|_| AppError::Unauthorized)?;
+
+        tracing::info!("password hash: {}", user.password_hash);
```
hanneskt: issue: also don't log password hashes

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,75 @@
...
+
+
+        // 2. Verify password
+        let parsed_hash = PasswordHash::new(&user.password_hash)
+            .map_err(|_| AppError::Unauthorized)?;
```
hanneskt: issue: use a different error here. Maybe Internal?

 if the password hash from the database can't be parsed, we have bigger problems as the hash in the database is invalid. A user can do nothing to fix this.

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        Argon2::default()
+            .verify_password(payload.password.as_bytes(), &parsed_hash)
+            .map_err(|_| AppError::Unauthorized)?;
+
+        // 3. Generate JWT
+        let exp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
```
hanneskt: suggestion: replace this unwrap with a `?` by adding an error to the `AppError` enum with a `#[from]` field

Although this can't really happen i guess, so probably fine like it is.

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,75 @@
...
+#[derive(Serialize)]
+pub struct AuthResponse {
+    pub token: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Claims {
+    pub sub: String,
+    pub email: String,
+    pub exp: usize,
```
hanneskt: suggestion: use a `u64` instead of `usize`

`as_secs()` always returns a `u64`. while definitely not going to happen, `usize` on a 32bit system is only 32 bits and will break. There is no reason to use a `usize` here

backend/database/src/repos/user.rs
```diff
@@ -15,21 +15,31 @@ impl<'a> UserRepo<'a> {
...
+            .await?
+            .ok_or(DatabaseError::NotFound)
+    }
+
     pub async fn create(&self, user: UserCreate) -> Result<User, DatabaseError> {
         sqlx::query_as(
             "
-            INSERT INTO \"user\" (username) VALUES ($1)
-            RETURNING id, username;
+            INSERT INTO \"users\" (username, email, password_hash) VALUES ($1, $2, $3)
```
hanneskt: nitpick: if the table is called users, you don't need to quote it 

also ends with a double semicolon

hanneskt (dismissed):

backend/src/handlers/auth.rs
Thoosje: Possible but might give attackers more info abt what is going wrong? Now in the whole auth flow the errors should be the same?

Thoosje:
How do we want to handle logout? Simply remove the JWT from the cookies? This will work locally but if the JWT token gets stolen it still can be used. Better would be to blacklist the token in a redis with a hartbeat endpoint I think but this would be more work and complex.

hanneskt:
> How do we want to handle logout? Simply remove the JWT from the cookies? This will work locally but if the JWT token gets stolen it still can be used. Better would be to blacklist the token in a redis with a hartbeat endpoint I think but this would be more work and complex.

I usually don't use JWTs at all, and just use session cookies with the data stored on the server. This allows you to easily wipe them.

But, something like these refresh tokens would work too: https://systemdesign.tech/post/revoking-jwts-a-simple-guide

We could also store a timestamp in the database and any tokens given out before that timestamp would not be accepted anymore. This requires storing the creation date in the jwt as well, and an extra field on the user. It can then be updated to invalidate all sessions.
That might be a good middle ground? It's not perfect but pretty simple and can be used in case of a leak.

Thoosje:
I hope nothing broke/will break. Had to refactor almost everything to get it working with the main branch...

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,259 @@
...
+use std::time::{SystemTime, UNIX_EPOCH};
+use time::Duration;
+use uuid::Uuid;
+
+use crate::{config::AppConfig, error::AppError, error::ErrorResponse, extractors::auth::AuthUser};
+use database::{Database, models::session::SessionCreate};
+use utoipa::ToSchema;
+
+const ACCESS_TOKEN_EXPIRY_MINUTES: i64 = 5;
+const REFRESH_TOKEN_EXPIRY_DAYS: i64 = 7;
```
tst-038: Consider placing these in ENV variables

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,259 @@
...
+        ("refresh_token" = [])
+    )
+)]
+pub async fn refresh(
+    jar: CookieJar,
+    db: Database,
+    config: AppConfig,
+) -> Result<(CookieJar, Json<AuthResponse>), AppError> {
+    let refresh_token = jar
+        .get("refresh_token")
```
tst-038, line 182: Use constants for these to prevent name inconsistency

backend/src/handlers/auth.rs
```diff
@@ -0,0 +1,259 @@
...
+        let _ = db.sessions().delete(user.session_id).await;
+    }
+
+    if let Some(cookie) = jar.get("refresh_token") {
+        let refresh_token = cookie.value();
+        let hashed_token = hash_token(refresh_token);
+        let _ = db.sessions().delete_by_token_hash(&hashed_token).await;
+    }
+
+    let access_cookie = Cookie::build(("access_token", ""))
```
tst-038, line 252: Use constants for these to prevent name inconsistency

backend/src/lib.rs
```diff
@@ -32,11 +36,31 @@ pub struct AppState {
...
 )]
 pub struct ApiDoc;
+struct SecurityAddon;
+
+impl Modify for SecurityAddon {
+    fn modify(&self, openapi: &mut utoipa::openapi::OpenApi) {
+        if let Some(components) = openapi.components.as_mut() {
+            // Register Access Token (JWT)
+            components.add_security_scheme(
+                "cookie_auth", // Internal name for the scheme
```
tst-038, line 53: Move to a constant

frontend/src/app/[locale]/admin/page.tsx
Comment (tst-038) [inline frontend/src/app/[locale]/admin/page.tsx:1]: Implement with i18n support, use seperate components like  AdminHeader, QuickActionsCard, AdminDashboard

frontend/src/components/auth/LoginForm.tsx
tst-038, line 1: Implement with i18n support, use seperate components like FormError, InputField.

frontend/src/components/auth/LoginForm.tsx
```diff
@@ -0,0 +1,67 @@
...
+            email: formData.get("email") as string,
+            password: formData.get("password") as string,
+        });
+    };
+
+    const axiosError = error as AxiosError;
+    const isUnauthorized = axiosError?.response?.status === 401;
+
+    return (
+        <div className="mx-auto flex w-full max-w-sm flex-col gap-6">
```
tst-038: Container `div (mx-auto max-w-sm flex flex-col gap-6)` can be a layout component (allowing for more consistency between login and possible future register or other similar auth pages.)

frontend/src/components/auth/LoginForm.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const { mutate, isPending, error, isError } = useLogin();
+
+    const handleSubmit = (e: React.FormEvent) => {
+        e.preventDefault();
+        const formData = new FormData(e.target as HTMLFormElement);
+        mutate({
+            email: formData.get("email") as string,
+            password: formData.get("password") as string,
+        });
+    };
```
tst-038: A safer approach would be to do
```
const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
    e.preventDefault();
    const form = e.currentTarget;
    const email = form.email.value;
    const password = form.password.value;

    mutate({ email, password });
};
```

frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,65 @@
...
+    const queryClient = useQueryClient();
+    const router = useRouter();
+
+    return useMutation({
+        mutationFn: async (credentials: LoginDTO) => {
+            const { data } = await api.post("/auth/login", credentials);
+            return data;
+        },
+        onSuccess: async () => {
+            await queryClient.refetchQueries({ queryKey: ["user"] });
```
tst-038: InvalidateQueries marks the query stale, allowing for auto refetching.
```typescript
queryClient.invalidateQueries({ queryKey: ["user"] });
```

frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,65 @@
...
+export const useLogout = () => {
+    const queryClient = useQueryClient();
+    const router = useRouter();
+
+    return useMutation({
+        mutationFn: async () => {
+            await api.post("/auth/logout");
+        },
+        onSuccess: () => {
+            queryClient.clear();
```
tst-038: This is quite inefficient, it deletes ALL queries resulting in irrelevant cached data to be deleted.
Use
```typescript
queryClient.removeQueries({ queryKey: ["user"] });
```

The query key should be made a constant.

frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,65 @@
...
+            router.push("/login");
+            toast.success("Logged out successfully");
+        },
+        onError: () => {
+            toast.error("An error occurred during logout");
+            // Clear local state anyway
+            queryClient.clear();
```
tst-038: See comment on line R54

frontend/src/lib/api.ts
```diff
@@ -0,0 +1,66 @@
...
+            originalRequest._retry = true;
+            isRefreshing = true;
+
+            try {
+                await api.post("/auth/refresh");
+                processQueue(null);
+                return api(originalRequest);
+            } catch (refreshError) {
+                processQueue(refreshError as AxiosError, null);
+                queryClient.clear();
```
tst-038:
```typescript
queryClient.removeQueries({ queryKey: ["user"] });
```

frontend/src/lib/api.ts
```diff
@@ -0,0 +1,66 @@
...
+    (response) => response,
+    async (error) => {
+        const originalRequest = error.config;
+
+        // Skip refresh for login/logout and if it's already a retry
+        if (
+            error.response?.status === 401 &&
+            !originalRequest._retry &&
+            !originalRequest.url?.includes("/login") &&
+            !originalRequest.url?.includes("/logout")
```
tst-038: This could cause trouble on "/admin/login-history" if it were to exist at some point.
```typescript
const isAuthRoute =
  originalRequest.url?.startsWith("/login") ||
  originalRequest.url?.startsWith("/logout")
```

frontend/src/types/auth.types.ts
```diff
@@ -0,0 +1,9 @@
+export interface LoginDTO {
+    email: string;
+    password: string;
+}
+
+export type AdminUser = {
+    user_id: string;
```
tst-038: `user_id`-> `id`

frontend/src/types/auth.types.ts
```diff
@@ -0,0 +1,9 @@
+export interface LoginDTO {
+    email: string;
+    password: string;
+}
+
+export type AdminUser = {
```
tst-038: `AdminUser`->`User`

frontend/src/types/dto/auth.types.ts
tst-038, line 1:
```
  types/
    api/ (Exact backend shape)
      auth.api.types.ts
    dto/ (to API)
      auth.dto.ts
    models/ (for UI)
      user.types.ts
```

with mapper in hooks for example `useUser` 
```typescript
function mapUser(response: UserResponse): User {
  return {
    id: response.user_id,
    email: response.email,
  };
}
```

tst-038 (changes requested):
Still requires i18n support, some small other changes are needed to (see comments)

frontend/src/proxy.ts
```diff
@@ -1,9 +1,29 @@
...
+        return NextResponse.redirect(new URL("/login", request.url));
+    }
+
+    if (isAuthRoute && token) {
+        return NextResponse.redirect(new URL("/admin", request.url));
+    }
+
+    // Process i18n routing if no auth redirects are needed
+    return intlMiddleware(request);
+}
```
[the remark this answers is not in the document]
Thoosje: Middleware.ts is in older versions of next, has been renamed to proxy.ts

https://nextjs.org/docs/app/api-reference/file-conventions/proxy

> Note: The middleware file convention is deprecated and has been renamed to proxy. See [Migration to Proxy](https://nextjs.org/docs/app/api-reference/file-conventions/proxy#migration-to-proxy) for more details.

frontend/src/lib/api.ts
Thoosje: Changed to correct paths:
- `/auth/login`
- `/auth/logout`

backend/src/lib.rs
Thoosje: Doesn't work with rust macros?

Thoosje (commented):
Perfect

tst-038 (approved):

PR #43, opened 2026-03-01
Import locations, halls, events from API

### LINKED ISSUE #41: Import locations, halls, events from api
Currently only the productions are imported. We need to at least import the following:

- locations
- halls
- events

Base yourself on how the productions are imported. Here's how it works:
- [The api importer is created and started in a separate thread.](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/src/lib.rs#L28) It holds a handle to the database, the token to connect to the api and a http client.
- [The thread calls `update_since_last()`](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/api/src/lib.rs#L68) which handles fetching the objects from the api for each type.
- The [update_productions](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/api/src/lib.rs#L127) function makes a stream over the paginated endpoint, every call to .next() gets the next page of the endpoint. It stops yielding when there are no more pages left.
- It inserts the results of all the pages in the database using [it's respective repo](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/database/src/repos/production.rs#L33)

To add support for a new endpoint:
- make [a create model](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/database/src/models/production.rs#L54) for your object
- add [a database repo](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/database/src/repos/production.rs#L9) for the model you're adding
- add it to [the Database struct ](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/database/src/lib.rs#L61)
- fill it with some query functions
- [implement From on your create model](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/api/src/models/production.rs#L59). Do this in the same file as the ApiModel. This defines how you convert from your api model to the database create model. If it can fail, implement `TryFrom` instead and handle failure later in your api update function. Look at the api endpoint schema's on [the API page](https://www.viernulvier.gent/api/docs#/Productions/api_v1productions_get_collection). you will probably have to extract the keys to the referenced objects from their paths and look up our Uuid's in the database to insert them as foreign keys.
    - for example, on productions 
      "media_gallery": "https://example.com/",
      "review_gallery": "https://example.com/",
      "poster_gallery": "https://example.com/",
      are paths to their endpoints. As we're importing everything, you only need to insert your current item with the correct foreign keys. You don't need to make recursive api calls.
- Make a new function that is a copy of `update_productions` but with a different endpoint and saving to it's own database table. handle failure of the api -> create model here.
- add your update function under [the one for productions](https://github.com/SELab-2/viernulvier-6/blob/d05b9eb71a1d542d01c7f2fbb91c0103c3136f35/backend/api/src/lib.rs#L75) make sure the order is correct for the relations in the database. make sure to call `await`.

brentvent:
DONE:
- implemented locations:
  - made name nullable by adding new up/down file to migrations. 'Name' field was not present in their API response, but did appear in the older CSV data.
  - did not know what to do with the Foreign Keys for spaces, the only referenced object inside locations
  - is the handling of is_owned_by_viernulvier robust enough?

- implemented halls:
  - halls table contains the ```UUID``` of the associated ```location```. API response for hall only contains a ```space```, which in turn contains a ```location```. Is this what we want? I currently put ```Uuid::nil()``` in that field.
 
Closes #41


backend/api/src/models/location.rs
```diff
@@ -0,0 +1,72 @@
...
+        let source_id = api
+            .id
+            .split('/')
+            .next_back() // removes and takes the end of the iterator, in this case the id
+            .and_then(|s| s.parse::<i32>().ok());
+        
+        let slug = match source_id { // FIX: leave it at that, since name is not present?
+            Some(id) => format!("location-{}", id),
+            None => "location-unknown".to_string(),
+        };
```
hanneskt: suggestion: maybe we should just remove the slug from the database for locations, since we can't really make a useful text slug?

if our slugs are just location-id, which will make our url look like `/locations/location-12` then we might as well use just the id I think.

backend/api/src/models/location.rs
```diff
@@ -0,0 +1,72 @@
...
+        
+        let slug = match source_id { // FIX: leave it at that, since name is not present?
+            Some(id) => format!("location-{}", id),
+            None => "location-unknown".to_string(),
+        }; 
+
+        let is_owned_by_viernulvier = match api.own_location.as_str() {
+            "1" => true,
+            _ => false,
+        };
```
hanneskt, line 47: suggestion: log unexpected values with `warn!`

Handle the expected false value explicitly: `"" => false` and have a separate case for unexpected values.
You can still return false when you encounter them, but also log a warning. You can put blocks in a match statement like so:
```
_ => {
	do_something();
	false
}
```
(Blocks ending with an expression, without a `;` will be implicitly returned)

backend/api/src/lib.rs
```diff
@@ -72,7 +75,12 @@ impl ApiImporter {
         let last_update_ts = self.get_last_updated().await;
 
         // TODO start transactions
-        let res = self.update_productions(&last_update_ts).await;
+        let res = async {
+            self.update_locations(&last_update_ts).await?;
+            self.update_productions(&last_update_ts).await?;
+            Ok::<(), reqwest::Error>(())
+        }
+        .await;
```
hanneskt, line 92: praise: nice!

backend/api/src/lib.rs
```diff
@@ -147,4 +155,28 @@ impl ApiImporter {
...
+            let locations = batch_result?;
+            let amt = locations.len();
+            info!("got {amt} locations from api");
+            for location in locations {
+                self.db
+                    .locations()
+                    .insert(location.into())
+                    .await
+                    .unwrap();
+            }
```
hanneskt, line 182: suggestion: I would do the id -> Uuid for spaces here

In the for loop, get the uuid of the space from the database, and then use that as the foreign key.
While maybe not the most "efficient" way, this wil work well and importing from the api is not a hot path.

hanneskt (commented):
so far so good :)

backend/migrations/20260304171633_drop_slug_from_locations.up.sql
```diff
@@ -0,0 +1,2 @@
+ATLER TABLE locations
```
hanneskt: issue: typo in `ATLER`

You can check if your migrations run by spinning up a database in a container. I use this command:
`docker run --name sel2-dev-postgres -p 5432:5432 -e POSTGRES_PASSWORD=<password> -e POSTGRES_USER=sel2-dev -e POSTGRES_DB=sel2-dev postgres:latest`

You can check the migrations then by
`sqlx migrate run && sqlx migrate revert --target-version 0 && sqlx migrate run`
This will check if they apply, revert and then apply again to make sure the down scripts also work

backend/api/src/models/hall.rs
```diff
@@ -0,0 +1,86 @@
...
+    models::localized_text::ApiLocalizedText,
+};
+
+use slug::slugify;
+use uuid::Uuid;
+
+#[derive(Debug, Deserialize)]
+pub struct ApiHall {
+    #[serde(rename = "@context")]
+    pub context: String,
```
hanneskt: issue: in the collection api we use, there is no field `@context`, so it crashes when trying to import it.

remove this field to fix it

backend/api/src/models/hall.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    pub seat_selection: String,
+    pub open_seating: String,
+
+    pub name: ApiLocalizedText,
+    pub remark: Option<ApiLocalizedText>, // did not find a single 'remark' field returned by the API
+
+    /// link to location where the space is located
+    /// "/api/v1/space/{id}"
+    pub space: String,
```
hanneskt: issue: apparently, some halls don't have spaces

example:
```json
{
  "@id": "/api/v1/halls/423",
  "@type": "Hall",
  "created_at": "2024-09-30T13:48:13+00:00",
  "updated_at": "2024-09-30T13:48:13+00:00",
  "seat_selection": "",
  "open_seating": "",
  "name": {
    "nl": "Wondelgemse Meersen",
    "en": "Wondelgemse Meersen"
  }
}
```
😭 

This means we need to handle this in the database also :(

Make sure to handle this properly when using the `extract_source_id` function as well

backend/api/src/helper.rs
```diff
@@ -13,3 +13,15 @@ pub fn flatten_loc(text: Option<ApiLocalizedText>) -> (Option<String>, Option<St
...
+/// Helper to extract the id out of a hyperlink that has the format
+/// "https://www.viernulvier.gent/api/v1/spaces/1". The last int gets extracted.
+pub fn extract_source_id(hyperlink: &str) -> i32 {
+    hyperlink
+        .rsplit('/')
+        .next()
+        .unwrap()
+        .parse()
+        .unwrap()
+}
```
hanneskt: praise: good use of a helper function

issue: if the string is not a url ending with a number (for example, it suddenly gets a trailing `/`, which is still the same resource, it will panic.

This panics the whole thread: no other imports will be done.

We should improve error handling throughout the whole api import crate, **but let's leave that for another PR** (see #62)

hanneskt (changes requested):

hanneskt:
and make sure to run `cargo fmt` at the end to make sure all your files are formatted :)
(we should have a precommit hook for this)

hanneskt (approved):

PR #66, opened 2026-03-06
OpenAPI voor de backend API

### LINKED ISSUE #53: OpenAPI voor de backend API
Dit gaan we doen met https://docs.rs/utoipa/latest/utoipa/

jarnex47:
- basis setup voor automatische SwaggerUI documentatie via utoipa-axum intergratie
- Docs en routing opgezet voor eerste twee endpoints: '/version' en '/productions', de rest zal nog volgen.

Refs OpenAPI voor de backend API
Fixes #53

### DIFF
--- a/backend/src/handlers/production.rs
+++ b/backend/src/handlers/production.rs
@@ -3,13 +3,6 @@ use database::Database;
 
 use crate::{dto::production::ProductionPayload, error::AppError};
 
-pub struct ProductionHandler;
-
-impl ProductionHandler {
-    pub async fn all(db: Database) -> Result<Json<Vec<ProductionPayload>>, AppError> {
-        Ok(Json(ProductionPayload::all(&db, 10).await?))
-    }
-}
 
 #[utoipa::path(
     method(get),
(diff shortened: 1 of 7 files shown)


backend/src/handlers/production.rs
```diff
@@ -10,3 +10,16 @@ impl ProductionHandler {
...
+    path = "/productions",
+    tag = "Productions",
+    description = "Get all productions",
+    responses(
+        (status = 200, description = "Success", body = [ProductionPayload])
+    )
+)]
+pub async fn all(db: Database) -> Result<Json<Vec<ProductionPayload>>, AppError> {
+    ProductionHandler::all(db).await
+}
```
hanneskt: suggestion: remove the `ProductionHandler` entirely

It's not really serving the purpose of grouping production routes anymore, and the separate `all` function just calls this one. It is fine to only have the function outside the `impl` block and remove the struct completely.
Just call the database directly in the outside `all` function.

hanneskt (approved):
Nice! Everything works and is implemented well.

If you fix this one thing, I would say we can merge this already, as the other routes will need to use this anyway, and it would be double work if we first implemented them the old way only to convert them to the OpenAPI way.

Issue #62, opened 2026-03-06
improve error handling throughout the api ingest crate


hanneskt:
Currently, during the api importing, if anything goes wrong, it just crashes and does not import anything else.

We should eliminate any `unwrap` and `expect` calls and choose what to do, so we are sure this does not happen. We'll need retries with backoffs, defaults and dropping where appropriate.

This will be quite involved, but important for the stability of our api import system as unwraps are not recoverable. Pass results, options and handle where needed instead.

We should keep track of failures, log them and make them visible in the CMS. My ideas currently:
- DB table with errors, can be quite free as most won't be really actionable for the admins anyway.
- Log warnings and errors, with enough context. This might mean the raw input json if it couldn't be parsed, possibly with a relevant section and the context of that section (maybe hard to implement)

PR #67, opened 2026-03-06
feat: events repo and import of events

jerjoen:
DONE:
- implemented events:
  - updated Event database model with ormlite
  - added EventRepo with by_id, all and insert
  - implemented From<ApiEvent> for EventCreate
  - added update_events, called after update_halls in update_since_last

### DIFF
--- a/backend/api/src/lib.rs
+++ b/backend/api/src/lib.rs
@@ -7,6 +7,7 @@ use reqwest::Client;
 use serde::de::DeserializeOwned;
 use tracing::{info, warn};
 
+use crate::models::event::ApiEvent;
 use crate::models::{
     collection::ApiCollection, hall::ApiHall, location::ApiLocation, production::ApiProduction,
     space::ApiSpace,
@@ -87,6 +88,7 @@ impl ApiImporter {
             self.update_spaces(&last_update_ts).await?;
             self.update_halls(&last_update_ts).await?;
             self.update_productions(&last_update_ts).await?;
+            self.update_events(&last_update_ts).await?;
             Ok::<(), reqwest::Error>(())
         }
         .await;
@@ -258,6 +260,25 @@ impl ApiImporter {
         }
 
         info!("Halls: finished importing");
+        Ok(())
+    }
+
+    pub async fn update_events(&self, updated_after: &str) -> Result<(), reqwest::Error> {
+        info!("start updating events");
+
+        let mut stream = pin!(self.paginated_collection::<ApiEvent>("/events", updated_after));
+
+        while let Some(batch_result) = stream.next().await {
+            let events = batch_result?;
+            let amt = events.len();
+            info!("got {amt} events from api");
+            for event in events {
+                self.db.events().insert(event.into()).await.unwrap();
+            }
+            info!("inserted {amt} events into db");
+        }
+
+        info!("finished importing events");
         Ok(())
     }
 }
--- a/backend/api/src/lib.rs
+++ b/backend/api/src/lib.rs
@@ -272,21 +272,20 @@ impl ApiImporter {
             let amt = events.len();
             info!("Events: got {amt} from api");
             for event in events {
-                let production_uuid = match event.production_source_id() {
+                let production_id = match event.production_source_id() {
                     Some(id) => {
-                        let uuid = self.db.productions().by_source_id(id).await
-                            .unwrap()
-                            .map(|p| p.id);
-                        if uuid.is_none() {
-                            warn!("Events: production source_id {id} not found in db, skipping");
+                        match self.db.productions().by_source_id(id).await.unwrap() {
+                            Some(p) => p.id,
+                            None => {
+                                warn!("Events: production source_id {id} not found in db, skipping");
+                                continue;
+                            }
                         }
-                        uuid
                     }
-                    None => None,
-                };
-
-                let Some(production_id) = production_uuid else {
-                    continue;
+                    None => {
+                        warn!("Events: event has no production source_id, skipping");
+                        continue;
+                    }
                 };
 
                 let hall_uuid = match event.hall_source_id() {
--- a/backend/api/src/lib.rs
+++ b/backend/api/src/lib.rs
@@ -272,9 +272,38 @@ impl ApiImporter {
             let amt = events.len();
             info!("Events: got {amt} from api");
             for event in events {
-                if let Err(e) = self.db.events().insert(event.into()).await {
-                    warn!("Events: failed to insert event: {e}");
-                }
+                let production_uuid = match event.production_source_id() {
+                    Some(id) => {
+                        let uuid = self.db.productions().by_source_id(id).await
+                            .unwrap()
+                            .map(|p| p.id);
+                        if uuid.is_none() {
+                            warn!("Events: production source_id {id} not found in db, skipping");
+                        }
+                        uuid
+                    }
+                    None => None,
+                };
+
+                let Some(production_id) = production_uuid else {
+                    continue;
+                };
+
+                let hall_uuid = match event.hall_source_id() {
+                    Some(id) => {
+                        let uuid = self.db.halls().by_source_id(id).await
+                            .unwrap()
+                            .map(|h| h.id);
+                        if uuid.is_none() {
+                            warn!("Events: hall source_id {id} not found in db");
+                        }
+                        uuid
+                    }
+                    None => None,
+                };
+
+                let event_create = event.to_create(production_id, hall_uuid);
+                self.db.events().insert(event_create).await.unwrap();
             }
             info!("Events: inserted {amt} into db");
         }
--- a/backend/api/src/lib.rs
+++ b/backend/api/src/lib.rs
@@ -7,13 +7,12 @@ use reqwest::Client;
 use serde::de::DeserializeOwned;
 use tracing::{info, warn};
 
-use crate::models::event::ApiEvent;
-use crate::models::{
-    collection::ApiCollection, hall::ApiHall, location::ApiLocation, production::ApiProduction,
-    space::ApiSpace,
-};
-
 use crate::helper::extract_source_id;
+use crate::models::space::ApiSpace;
+use crate::models::{
+    collection::ApiCollection, event::ApiEvent, hall::ApiHall, location::ApiLocation,
+    production::ApiProduction,
+};
 
 mod helper;
 pub mod models {
--- a/backend/api/src/lib.rs
+++ b/backend/api/src/lib.rs
@@ -263,21 +263,21 @@ impl ApiImporter {
     }
 
     pub async fn update_events(&self, updated_after: &str) -> Result<(), reqwest::Error> {
-        info!("start updating events");
+        info!("Events: start updating");
 
         let mut stream = pin!(self.paginated_collection::<ApiEvent>("/events", updated_after));
 
         while let Some(batch_result) = stream.next().await {
             let events = batch_result?;
             let amt = events.len();
-            info!("got {amt} events from api");
+            info!("Events: got {amt} from api");
             for event in events {
                 self.db.events().insert(event.into()).await.unwrap();
             }
-            info!("inserted {amt} events into db");
+            info!("Events: inserted {amt} into db");
         }
 
-        info!("finished importing events");
+        info!("Events: finished importing");
         Ok(())
     }
 }
--- a/backend/api/src/lib.rs
+++ b/backend/api/src/lib.rs
@@ -272,7 +272,9 @@ impl ApiImporter {
             let amt = events.len();
             info!("Events: got {amt} from api");
             for event in events {
-                self.db.events().insert(event.into()).await.unwrap();
+                if let Err(e) = self.db.events().insert(event.into()).await {
+                    warn!("Events: failed to insert event: {e}");
+                }
             }
             info!("Events: inserted {amt} into db");
         }
--- a/backend/api/src/models/event.rs
+++ b/backend/api/src/models/event.rs
@@ -63,9 +63,8 @@ impl ApiEvent {
             vendor_id: self.vendor_id,
             box_office_id: self.box_office_id,
             uitdatabank_id: self.uitdatabank_id,
-            max_tickets_per_order: self
-                .max_tickets_per_order
-                .map(|v| i32::try_from(v).expect("max_tickets_per_order out of range")),
+            max_tickets_per_order: self.max_tickets_per_order
+                .map(|v| v.try_into().expect("max_tickets_per_order out of range")),
             production_id,
             status: self.status,
             hall_id,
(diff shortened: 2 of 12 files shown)


Kaya-Sem:
wanneer klaar voor review?

jerjoen:
normaalgezien was alles al in orde (denk ik), maar vergeten klaar zetten voor review, my bad...

hanneskt (changes requested):
You'll need to change the `ApiEvent` struct, as currently it does not deserialize and no events are imported at all.
The `production` field is a map, not just a string. I understand I added the struct, but I couldn't test it without the api code.

It's probably easiest to test by commenting out [these lines](https://github.com/SELab-2/viernulvier-6/blob/af968a9e26c88e047a881471841f687308004530/backend/api/src/lib.rs#L90)

hanneskt (changes requested):
(oops too early)

backend/api/src/models/event.rs
```diff
@@ -26,12 +47,57 @@ pub struct ApiEvent {
...
-    pub hall: String,
-    pub prices: Vec<String>,
+    pub production: ApiEventProduction,
+    pub status: ApiEventStatus,
+    pub hall: ApiEventHall,
 
     pub info: ApiLocalizedText,
     pub eticket_info: ApiLocalizedText,
     pub external_order_url: ApiLocalizedText,
 }
```
hanneskt, line 43: We're using their [event collections endpoint](https://www.viernulvier.gent/api/docs#/Events/api_v1events_get_collection), which is a bit different than their specific [events endpoint](https://www.viernulvier.gent/api/docs#/Events/api_v1events_id_get).

As you can see in the schema of the event collections endpoint on their docs site, the members of the collections (which are the events here) are supposed to look like this:
```json
...
    {
      "@context": "string",
      "@id": "string",
      "@type": "string",
      "created_at": "2026-03-12T02:20:18.567Z",
      "updated_at": "2026-03-12T02:20:18.567Z",
      "starts_at": "2026-03-12T02:20:18.567Z",
      "ends_at": "2026-03-12T02:20:18.567Z",
      "intermission_at": "2026-03-12T02:20:18.567Z",
      "doors_at": "2026-03-12T02:20:18.567Z",
      "box_office_id": "string",
      "vendor_id": "string",
      "max_tickets_per_order": 0,
      "uitdatabank_id": "string",
      "secure": true,
      "sms_verification": true,
      "production": {},
      "status": "https://example.com/",
      "hall": "https://example.com/",
      "prices": [
        "https://example.com/"
      ],
      "info": {
        "additionalProp1": "string",
        "additionalProp2": "string",
        "additionalProp3": "string"
      },
      "eticket_info": {
        "additionalProp1": "string",
        "additionalProp2": "string",
        "additionalProp3": "string"
      },
      "external_order_url": {
        "additionalProp1": "string",
        "additionalProp2": "string",
        "additionalProp3": "string"
      }
    }
...
```
But if you actually try to query the endpoint, you get responses with members like:
```json
...
{
      "@id": "/api/v1/events/1",
      "@type": "Event",
      "created_at": "2014-07-01T08:07:04+00:00",
      "updated_at": "2022-10-03T07:42:31+00:00",
      "starts_at": "2017-11-20T13:30:00+00:00",
      "box_office_id": "",
      "vendor_id": "",
      "max_tickets_per_order": 0,
      "secure": false,
      "sms_verification": false,
      "production": {
        "@type": "StandardProduction",
        "@id": "/api/v1/productions/1890"
      },
      "status": "/api/v1/events/statuses/1",
      "hall": "/api/v1/halls/1",
      "prices": [],
      "external_order_url": {
        "nl": "https://www.google.nl/"
      }
    }
...
```

As you can see, they never return the `"@context"` key, so you need to drop that from your model.
Second, the `"status"` is here not nested, it's just a string, so you need to adjust that. `ApiEventStatus` is thus not needed.
`"hall"` is similar, it's not nested here and thus just needs to be deserialized as a string and `ApiEventHall` is not needed either.

After fixing that and running it against the api you notice that some fields aren't always returned (and are thus optional). Our model needs to reflect that by having their fields be `Option<T>` where `T` is their original type.
`ends_at`, `max_tickets_per_order`, `uitdatabank_id`, `info`, `eticket_info` and `external_order_url` are all optional. I found this through trial and error.

You'll need to make the corresponding fields of the database model optional as well so you can just pass them and they will end up saved as `null`.

backend/api/src/models/event.rs
```diff
@@ -26,12 +47,57 @@ pub struct ApiEvent {
...
+
+fn parse_uuid_from_iri(iri: &str) -> Option<Uuid> {
+    let trimmed = iri.trim_end_matches('/');
+    trimmed
+        .rsplit('/')
+        .next()
+        .filter(|segment| !segment.is_empty())
+        .and_then(|segment| Uuid::parse_str(segment).ok())
+}
```
hanneskt: Naming it `iri` is a bit confusing, as I don't think many people are familiar with [Internationalized Resource Identifiers](https://en.wikipedia.org/wiki/Internationalized_Resource_Identifier), which I assume is meant here?

Then, as you can see in the api response:
```json
...
"production": {
        "@type": "StandardProduction",
        "@id": "/api/v1/productions/1890"
      },
      "status": "/api/v1/events/statuses/1",
      "hall": "/api/v1/halls/1",
...
```
Their api uses numerical id's, not UUIDs so this function always fails.

We actually already have [a helper function](https://github.com/SELab-2/viernulvier-6/blob/0b2de2737b936c9fcebb516e75f132cadef4a4e2/backend/api/src/helper.rs#L19-L31) for exactly this which is already used in other api import models. Use that instead.

The filter seems a bit redundant as you already trim the ends of extra `/`, but I guess it can't hurt to be safe?

backend/api/src/models/event.rs
```diff
@@ -26,12 +47,57 @@ pub struct ApiEvent {
...
+impl From<ApiEvent> for EventCreate {
+    fn from(api: ApiEvent) -> Self {
+        let production_id = match parse_uuid_from_iri(&api.production.id) {
+            Some(id) => id,
+            None => {
+                eprintln!("warning: invalid production id in event: {}", api.production.id);
+                Uuid::nil()
+            }
+        };
+
```
hanneskt: Use the [`extract_source_id`](https://github.com/SELab-2/viernulvier-6/blob/0b2de2737b936c9fcebb516e75f132cadef4a4e2/backend/api/src/helper.rs#L21) function instead.

As we use tracing, use `warn!()` instead of `eprintln`. It works with our existing logging and gives a lot more information.

This is not relevant as these aren't UUIDs, but if this was the case, returning a fully zero uuid as a recovery from failure seems dangerous. You create the illusion of there being something relevant (a UUID) but actually it's not usable at all as it's just zeroed. We would need a special check on zeroed UUIDs to hande this.
If this was the case, the better way to handle this would be to make the database field nullable and signal a missing link that way.

You would have to check the api, but I hope this field is always set. Otherwise you will have to make this optional as well, just like in [halls](https://github.com/SELab-2/viernulvier-6/blob/0b2de2737b936c9fcebb516e75f132cadef4a4e2/backend/database/src/models/hall.rs#L21) for example.

Just a tip for other parts of the code: If you need a fallback like this, you can use `unwrap_or_else` like this:
```rust
let hall_id = parse_uuid_from_iri(&api.hall).unwrap_or_else(|| {
    eprintln!("warning: invalid hall id in event: {}", api.hall);
    Uuid::nil()
});
```

backend/api/src/models/event.rs
```diff
@@ -26,12 +47,57 @@ pub struct ApiEvent {
...
+        let hall_id = match parse_uuid_from_iri(&api.hall.id) {
+            Some(id) => id,
+            None => {
+                eprintln!("warning: invalid hall id in event: {}", api.hall.id);
+                Uuid::nil()
+            }
+        };
```
hanneskt: Same as [this comment](https://github.com/SELab-2/viernulvier-6/pull/67/changes/BASE..0b2de2737b936c9fcebb516e75f132cadef4a4e2#r2921961501)

backend/api/src/models/event.rs
```diff
@@ -26,12 +47,57 @@ pub struct ApiEvent {
...
+            updated_at: api.updated_at,
+            started_at: api.starts_at,
+            ended_at: api.ends_at,
+            intermission_at: api.intermission_at,
+            doors_at: api.doors_at,
+            vendor_id: api.vendor_id,
+            box_office_id: api.box_office_id,
+            uitdatabank_id: api.uitdatabank_id,
+            max_tickets_per_order: i32::try_from(api.max_tickets_per_order)
+                .expect("max_tickets_per_order out of range for i32"),
```
hanneskt: Use `api.max_tickets_per_order.try_into()` instead.

Shouldn't use expect, but this is okay since we need to think about error handling and we will fix this for #62

backend/api/src/lib.rs
```diff
@@ -260,4 +261,25 @@ impl ApiImporter {
...
+        let mut stream = pin!(self.paginated_collection::<ApiEvent>("/events", updated_after));
+
+        while let Some(batch_result) = stream.next().await {
+            let events = batch_result?;
+            let amt = events.len();
+            info!("Events: got {amt} from api");
+            for event in events {
+                if let Err(e) = self.db.events().insert(event.into()).await {
+                    warn!("Events: failed to insert event: {e}");
+                }
```
hanneskt: Dropping events like this is dangerous, as they will never be reimported and will be permanently lost.

This is something we need to rethink, but I would rather hard error here then just log a warning and end up having an inconsistent database. We will improve error handling later.

backend/database/src/models/event.rs
```diff
@@ -1,14 +1,16 @@
...
+#[ormlite(insert = "EventCreate")]
+#[ormlite(table = "events")]
 pub struct Event {
     pub id: Uuid,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
-    pub starts_at: DateTime<Utc>,
-    pub ends_at: DateTime<Utc>,
+    pub started_at: DateTime<Utc>,
+    pub ended_at: DateTime<Utc>,
```
hanneskt: it seems arbitrary to have these be different than their api to put them in the past tense. Some events we import will be in the future.

I could see maybe naming them `start_time` and `end_time`?
I would say either keep them like in the api, or improve the naming, but making it past tense does not do that.

backend/api/src/lib.rs
```diff
@@ -260,4 +261,25 @@ impl ApiImporter {
...
+    pub async fn update_events(&self, updated_after: &str) -> Result<(), reqwest::Error> {
+        info!("Events: start updating");
+
+        let mut stream = pin!(self.paginated_collection::<ApiEvent>("/events", updated_after));
+
+        while let Some(batch_result) = stream.next().await {
+            let events = batch_result?;
+            let amt = events.len();
+            info!("Events: got {amt} from api");
+            for event in events {
```
hanneskt, line 274: Here you will have to look for the UUID for any linked resources, like the hall and the production. You can do this [just like halls](https://github.com/SELab-2/viernulvier-6/blob/0b2de2737b936c9fcebb516e75f132cadef4a4e2/backend/api/src/lib.rs#L233-L256) for example.

You'll need to create a function to convert it with the extra UUIDs to a database model like [this one](https://github.com/SELab-2/viernulvier-6/blob/0b2de2737b936c9fcebb516e75f132cadef4a4e2/backend/api/src/models/hall.rs#L41).

hanneskt (changes requested):
This still doesn't run. Do you need instructions on how to run this against their API so you can test if your code works?

I included most things needed to fix this, but you will need to check yourself if it can ingest all events from their api by running it locally.

backend/database/src/models/event.rs
jerjoen: Copilot gaf gisteren deze opmerking:
> With `#[derive(Model)]` and `#[ormlite(table = "events")]`, ormlite will map struct fields to DB column names by default. The `events` table schema uses `started_at`/`ended_at`, but this struct has `starts_at`/`ends_at`, so generated SQL will reference columns that don’t exist and inserts/selects will fail. Rename the fields or add explicit column mapping attributes for those columns (and ensure the generated `EventCreate` matches).

Ik heb dit zelf bij het testen ook opgemerkt, en is dus waarom ik het zo heb laten staan. Ik heb wel `ended_at` optional gemaakt.

jerjoen:
Heb lokaal getest. Events worden correct gedeserialiseerd en de import start, maar is een crash bij halls.

backend/api/src/models/event.rs
```diff
@@ -15,23 +22,51 @@ pub struct ApiEvent {
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
     pub starts_at: DateTime<Utc>,
-    pub ends_at: DateTime<Utc>,
+    pub ends_at: Option<DateTime<Utc>>,
```
jarnex47, line 25: `ends_at` staat als `NOT NULL` in de migration, maar is `Option<DateTime<Utc>>` in het model.  
Moet deze kolom nullable zijn?

backend/api/src/models/event.rs
```diff
@@ -15,23 +22,52 @@ pub struct ApiEvent {
...
+            updated_at: self.updated_at,
+            started_at: self.starts_at,
+            ended_at: self.ends_at,
+            intermission_at: self.intermission_at,
+            doors_at: self.doors_at,    
+            vendor_id: self.vendor_id,
+            box_office_id: self.box_office_id,
+            uitdatabank_id: self.uitdatabank_id,
+            max_tickets_per_order: self.max_tickets_per_order
+                .map(|v| i32::try_from(v).expect("max_tickets_per_order out of range")),
```
hanneskt: suggestion: use `try_into()` here instead

```rust
v.try_into().expect("max_tickets_per_order out of range")
```

backend/api/src/lib.rs
```diff
@@ -260,4 +261,54 @@ impl ApiImporter {
...
+                            warn!("Events: production source_id {id} not found in db, skipping");
+                        }
+                        uuid
+                    }
+                    None => None,
+                };
+
+                let Some(production_id) = production_uuid else {
+                    continue;
+                };
```
hanneskt: question: why do you continue here? 

If there is no production id, do we just not add the event to the database? That's fine, but a weird thing to get from the database, so we should at least log it with a `warn!`.

suggestion: If you do keep the continue, integrate it into the match above. A `if let` block with an else case might be cleaner in that case

hanneskt (changes requested):
A couple changes but looks good otherwise

hanneskt (approved):

jarnex47 (approved):
tested locally, everyhting works fine

Kaya-Sem (approved):

PR #90, opened 2026-03-10
locations get, production get, delete, update, create endpoints

jarnex47:


hanneskt (approved):
Looks good!

backend/src/handlers/location.rs
```diff
@@ -0,0 +1,37 @@
...
+};
+
+#[utoipa::path(
+    method(get),
+    path = "/locations",
+    tag = "Locations",
+    description = "Get all locations",
+    responses(
+        (status = 200, description = "Success", body = [LocationPayload])
+    )
```
Thoosje, line 17: Error codes are needed I think: probably only 5xx here

backend/src/handlers/location.rs
```diff
@@ -0,0 +1,37 @@
...
+    method(get),
+    path = "/locations/{id}",
+    tag = "Locations",
+    description = "Get location by id",
+    params(
+        ("id" = Uuid, Path, description = "Location UUID")
+    ),
+    responses(
+        (status = 200, description = "Success", body = LocationPayload)
+    )
```
Thoosje, line 34: Error codes are needed I think: 5xx and 404 when location by id does not exist

backend/src/handlers/production.rs
```diff
@@ -13,6 +16,71 @@ use crate::{dto::production::ProductionPayload, error::AppError};
...
+    method(get),
+    path = "/productions/{id}",
+    tag = "Productions",
+    description = "Get a production by id",
+    params(
+        ("id" = Uuid, Path, description = "Production UUID")
+    ),
+    responses(
+        (status = 200, description = "Success", body = ProductionPayload)
+    )
```
Thoosje, line 34: Error codes are needed I think: 5xx and 404

backend/src/handlers/production.rs
```diff
@@ -13,6 +16,71 @@ use crate::{dto::production::ProductionPayload, error::AppError};
...
+}
+
+#[utoipa::path(
+    method(post),
+    path = "/productions",
+    tag = "Productions",
+    description = "Create a production",
+    responses(
+        (status = 201, description = "Created", body = ProductionPayload)
+    )
```
Thoosje, line 47: Error codes are needed I think: 5xx

backend/src/handlers/production.rs
```diff
@@ -13,6 +16,71 @@ use crate::{dto::production::ProductionPayload, error::AppError};
...
+}
+
+#[utoipa::path(
+    method(put),
+    path = "/productions",
+    tag = "Productions",
+    description = "Update the fields of a production",
+    responses(
+        (status = 200, description = "Success", body = ProductionPayload)
+    )
```
Thoosje, line 82: Error codes are needed I think: 5xx and 404 if production is not found. Maybe also error code when incorrect fields are sent

Thoosje (changes requested):

backend/src/handlers/location.rs
hanneskt: I think 500 are always possible (but never expected) so they should not be added to the docs as a possibility

backend/src/dto/location.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub code: Option<String>,
+    pub street: Option<String>,
+    pub number: Option<String>,
+    pub postal_code: Option<String>,
+    pub city: Option<String>,
+    pub country: Option<String>,
+    pub phone_1: Option<String>,
+    pub phone_2: Option<String>,
+    pub is_owned_by_viernulvier: Option<bool>,
+    pub uitdatabank_id: Option<String>,
```
Thoosje, line 26: Maybe rename this to `original_id`?

backend/src/dto/production.rs
```diff
@@ -54,14 +87,48 @@ pub struct ProductionPayload {
...
+    pub info_nl: Option<String>,
+    pub info_en: Option<String>,
+    pub description_short_nl: Option<String>,
+    pub description_short_en: Option<String>,
+    pub eticket_info: Option<String>,
+
+    // pub genres: Option<String>,
+    // pub uitdatabank_keywords: Option<String>,
+    pub uitdatabank_theme: Option<String>,
+    pub uitdatabank_type: Option<String>,
```
Thoosje, line 133: Maybe rename these also to `original_theme` and `original_id`?

Thoosje (approved):

backend/src/dto/location.rs
hanneskt: uitdatabank comes from their api, and is an external thing, so this should be kept. We keep track of their api id in `soucre_id`

backend/src/dto/production.rs
hanneskt: same as above

PR #97, opened 2026-03-12
add backend README

hanneskt:


brentvent (approved):
Optional: add docker instructions: ``` docker run --name sel2-dev-postgres -e POSTGRES_USER=sel2-dev -e POSTGRES_PASSWORD=password -e POSTGRES_DB=sel2-dev -p 5432:5432 -d postgres```, but this is basically the same for podman anyway

PR #103, opened 2026-03-14
fix: rename events start and end time naming

Kaya-Sem:
alters fields ni #67


tst-038:
What's the reasoning behind this change exactly? I'm assuming their api just returns the time only?

Kaya-Sem:
> What's the reasoning behind this change exactly? I'm assuming their api just returns the time only?

see https://github.com/SELab-2/viernulvier-6/pull/67#discussion_r2921984680

tst-038:
> > What's the reasoning behind this change exactly? I'm assuming their api just returns the time only?
> 
> see [#67 (comment)](https://github.com/SELab-2/viernulvier-6/pull/67#discussion_r2921984680)

Hmm, that's a valid point, however something like `start_at` or `starts_at` seems a better approach to me. Uses the well known _at conventions and doesn't allow for confusion regarding the time being perceived as a literal time-of-day.

Kaya-Sem:
> > > What's the reasoning behind this change exactly? I'm assuming their api just returns the time only?
> > 
> > 
> > see [#67 (comment)](https://github.com/SELab-2/viernulvier-6/pull/67#discussion_r2921984680)
> 
> Hmm, that's a valid point, however something like `start_at` or `starts_at` seems a better approach to me. Uses the well known _at conventions and doesn't allow for confusion regarding the time being perceived as a literal time-of-day.

Lets stick to their naming and indeed use `starts_at` and `ends_at`

tst-038 (approved):

Issue #104, opened 2026-03-14
Tagging and relations


Kaya-Sem:
**Suggested tags and relations structure and its implementation**

I suggest we separate tags and entity relations, as they are different things and should be modeled separately.

- Facet: a category of classification. Like a drawer label
- Tag: a value within a facet. the thing inside the drawer
- tagging: attachting a tag to a specific entity (join record)

So if you have a production called Rosas danst Rosas, it has two tagging rows: one pointing to the tag "Dance", one to "World premiere". Those tags belong to facets "Discipline" and "Format".

## Faceted tags
describe the nature of an entity (its themes, disciplines, period, mood, audience). These are the metadata you browse and filter by.

#### taxonomy
discipline: theathre, dance, music, visual art, film, opera, performance art, installation
format: world premiere, co-production, residency, workshop, touring, commision, revival, festival
theme: identity, memory, politics, ecology, diaspora
audience: all ages, children, adult, professional
...

-> not all facets apply to all entities!

#### implementation

`OR` within facet (broadening), and `AND` across facets (narrowing).
`theme=memory&theme=identity` returns productions tagged with memory *or* identity, but `theme=memory&discipline=dance` returns productions that are *both* dance *and* about memory.

Faceted filters  rendered as a sidebar of collapsible groups, one per facet. Each value shows its count, selected values are checked/highlighted, and selecting one updates the URL and re-fetches. If count = 0, grey out.

- counts update as you filter. When you select `dance`, all other facet counts recalculate to reflect only what's possible within that selection. This prevents dead-end states where a user selects a combination that returns zero results.
- **zero count values grey out.** if no dance production has format revival, that option becomes unselectable when dance is active. Rather than dissapearing entirely (disorienting) or staying clickable (zero results).
- **active filters appear as chips* The currently selected filters are summarised above the results so the user always knows whats active and can remove individual filters without hunting back to the sidebar.
- **URL reflects state**. each filter updates the URL (`?discipline=dance&theme=memory`). filters are bookmarkable, shareable, ... this is an archive!

# Relations
Too smelly to implement with tags only. relations carry their own data. the link between an artist and a production has a *role* (`director`, `performer`, `set designer`). That role belongs to neither the artist nor the production. It belongs to the edge between them. These relations are also bidirectional.

## Where they overlap
`tag` `discipline:dance` on a production -> describes the production itself, no linked entity
`relation` `an artist linked to a production` -> the artist is a first class entity with a profile
[...]

Issue #91, opened 2026-03-10
chore(backend): Standardize DTO Naming Across the Codebase


tst-038:
## Problem

Many DTOs currently use vague or inconsistent names like `ProductionPayload`. This makes it unclear whether a struct is for requests, responses, or internal use. The same problem likely exists for other entities, not just `Production`.

## Proposed Solution

Adopt clear, action-oriented, and consistent naming for all DTOs:

- `CreateXRequest` – for creation requests  
- `UpdateXRequest` – for update requests  
- `XResponse` – for API responses  

Optionally, append `Dto` if the folder structure does not already indicate these are DTOs:

- `CreateProductionDto`  
- `UpdateProductionDto`  
- `ProductionResponseDto`  

Refactor all existing DTOs to follow this convention.

PR #105, opened 2026-03-14
feat: tagging

Kaya-Sem:
Implements #104 

Introduces facets and tags into the database and rust backend.
Added taxonomy endpoint to retrieve facets and their tags


backend/migrations/20260314123706_tagging.down.sql
```diff
@@ -0,0 +1,7 @@
+DROP VIEW IF EXISTS entity_tags;
+DROP VIEW IF EXISTS tags_with_facet;
+
+DROP TABLE IF EXISTS taggings;
+DROP TABLE IF EXISTS tags;
+DROP TABLE IF EXISTS facet_entity_types;
+DROP TABLE IF EXISTS facets;
```
hanneskt: suggestion: remove the `IF EXISTS` from the drops

I think it's better that the downwards migrations fail if there is something wrong (as that is a sign that there is a problem), instead of silently doing nothing.

backend/src/dto/facet.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[derive(Serialize, ToSchema)]
+pub struct TagDto {
+    pub slug: String,
+    pub label: String,
+    pub sort_order: i32,
+}
+
+#[derive(Serialize, ToSchema)]
+pub struct FacetDto {
```
hanneskt: These namings break our current convention, but I am open to changing the convention.

There are some suggestions in #91.

docs/.gitignore
hanneskt, line 1: what is this doing here?

backend/database/src/repos/tag.rs
```diff
@@ -0,0 +1,93 @@
...
+            WHERE entity_type = $1 AND entity_id = $2
+            "#,
+            entity_type,
+            entity_id
+        )
+        .fetch_all(self.db)
+        .await?)
+    }
+
+    pub async fn facets(
```
hanneskt: suggestion: split this function into 2. `with_facets()` and `with_facets_for_entity(entitiy)`

Basically no code duplication and will make the usage these functions a lot more readable. And you don't need to add `None` when you want all tags.

```rust
db.tags().with_facets();
db.tags().with_facets_for_entity(');
```

backend/database/src/repos/tag.rs
```diff
@@ -0,0 +1,93 @@
...
+                r#"
+                SELECT
+                    f.slug       AS facet_slug,
+                    f.label      AS facet_label,
+                    f.sort_order AS facet_sort_order,
+                    t.slug       AS tag_slug,
+                    t.label      AS tag_label,
+                    t.sort_order AS tag_sort_order
+                FROM tags t
+                JOIN facets f ON f.id = t.facet_id
```
hanneskt: question: You have this query as a view, why not use that? (or why have the view?)

backend/migrations/20260314123706_tagging.up.sql
```diff
@@ -0,0 +1,175 @@
+-- Arts archive — tags schema
+
+--   - "facet" is the namespace (discipline, theme, format, …)
+--   - "value" is the human label ("World premiere")
+--   - "slug"  is the URL-safe version ("world-premiere")
+
+
+-- facets: defines the available facet namespaces and which entity types they apply to.
+CREATE TABLE facets (
```
hanneskt: question: are the facets every meant to be changed (from the cms for example)?

backend/migrations/20260314123706_tagging.down.sql
Kaya-Sem: Geeft dit geen problemen met ordering?

docs/.gitignore
Kaya-Sem: git ignoring. It is normal to have more than one git ignore file. can move if needed.

hanneskt: It was more why was a gitignore for the docs added in the tagging branch :p

backend/migrations/20260314123706_tagging.down.sql
hanneskt: Wat bedoel je met ordering? Migrations worden in volgorde van de timestamp uitgevoerd, en de files van boven naar onder, dus als je dingen dropt in de omgekeerde volgorde van ze toe te voegen dan is dat goed

backend/migrations/20260314123706_tagging.up.sql
Kaya-Sem: as in, the name of a facet? creating extra facets? changing values within facets?

backend/database/src/repos/tag.rs
Kaya-Sem: The view and query don't exactly line up, but the existence of the view is indeed debatable. Anticipation for future uses

docs/.gitignore
Kaya-Sem: think I tested some frontend build or so for the API integration, woops

backend/src/handlers/taxonomy.rs
```diff
@@ -0,0 +1,35 @@
...
+    )
+)]
+pub async fn get_facets(
+    db: Database,
+    Query(params): Query<TaxonomyParams>,
+) -> JsonResponse<Vec<FacetDto>> {
+    FacetDto::all(&db, params.entity_type.as_deref())
+        .await?
+        .json()
+}
```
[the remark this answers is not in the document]
Kaya-Sem: fix: resolved in #109

docs/.gitignore
Kaya-Sem: fix: is fixed in fumadocs pr #59

backend/migrations/20260314123706_tagging.up.sql
Kaya-Sem: see also #125

Kaya-Sem:
note: enum reimplement #144

backend/migrations/20260314123706_tagging.up.sql
Kaya-Sem: resolved: switched to enums

hanneskt (approved):

backend/README.md
```diff
@@ -51,3 +51,15 @@ cargo test --all
...
+
+## sqlx offline mode
+
+The Docker build uses `SQLX_OFFLINE=true`, so sqlx checks queries at compile time using the cached metadata in `.sqlx/` instead of connecting to a live database.
+
+Whenever you add or modify a `query!` or `query_as!` macro, regenerate the cache and commit it:
+
+```sh
+cargo sqlx prepare --workspace
+```
```
Thoosje, line 63: Add to husky so this gets done automatically?

backend/database/src/models/tag.rs
```diff
@@ -0,0 +1,38 @@
...
+}
+
+#[derive(Debug, FromRow)]
+pub struct TagWithFacet {
+    pub facet_slug: String,
+    pub facet_label: String,
+    pub tag_slug: String,
+    pub tag_label: String,
+    pub tag_sort_order: i32,
+}
```
Thoosje, line 41: Could you clarify the differences between the 3 type of tags. As the `Tag` also has a `facet` like the `TagWithFacet`. Also how does the `EntityTag` differ from the `Tagging`?

backend/database/src/models/tagging.rs
```diff
@@ -0,0 +1,15 @@
...
+#[derive(Debug, Model)]
+#[ormlite(insert = "TaggingCreate")]
+#[ormlite(table = "taggings")]
+pub struct Tagging {
+    pub id: Uuid,
+    pub tag_id: Uuid,
+    pub entity_type: EntityType,
+    pub entity_id: Uuid,
+    pub inherited: bool,
+}
```
Thoosje, line 16: How does the `EntityTag` differ from the `Tagging`?

backend/README.md
Kaya-Sem: this sounds like a good idea, but i know I would find it annoying quite quickly, needing a running db just to push.

backend/database/src/models/tag.rs
Kaya-Sem: 'tagging' is the raw table row, which is the write model. An 'EntityTag' was made to answer the question 'what tags does this entity have'. The 'TagWithFacet' was made for "what tags exist in the taxonomy". I agree that the naming might not be clear.

Merging these two is not straightforward, as the query maps column for column, and would degrade useage experience. I would suggest renaming:

'TagWithFacet' -> 'TaxonomyTag' (vocabulary entry, not entity bound)
'EntityTag' either stays or renamed to 'BoundTag', 'AttachedTag', ... (read projection, entity scoped!)

backend/database/src/models/tagging.rs
Kaya-Sem: It would be used as a write model (insert/update a row). EntityTag is the read model (query the view for display). They map to different things in the db. Tagging is not used for now, and we have no endpoints or ways to create new tags

backend/database/src/models/tag.rs
Thoosje: If you think it's necessary to rename it you can. Otherwise a comment near these types might help others reading the types.

Kaya-Sem: Renamed to `TaxonomyTag`.

backend/src/lib.rs
```diff
@@ -201,6 +201,10 @@ fn editor_routes(state: AppState) -> OpenApiRouter<AppState> {
         .routes(routes!(space::post))
         .routes(routes!(space::delete))
         .routes(routes!(space::put))
+  
+        // taxonomies
+        .routes(routes!(taxonomy::get_facets))
+
         .layer(from_extractor_with_state::<EditorUser, AppState>(state))
 }
```
Thoosje, line 209: Should this be protected?

Thoosje (approved):

PR #113, opened 2026-03-15
Test/location space hall endpoints

### LINKED ISSUE #114: Tests for API Endpoints
## API Endpoints

Track implementation and integration test coverage for all endpoints.
- [x] `Locations`
- [x] `Halls`
- [x] `Spaces`
- [x] `Productions`
- [ ] `Events`
- [ ] `Artists`
- [ ] `Blogposts`
- [ ] `Collections`
- [x] `Auth`
- [x] `Admin`

jarnex47:
- Fix repositories for locations, spaces, and halls to return **404** when deleting a non-existent resource
- Add **Serialize** on **PostPayload** types for location, space, and hall DTOs (needed for test helpers)
- Add CRUD integration tests for the **locations, spaces**, and **halls** endpoints, covering success and not-found cases for GET, POST, PUT, and DELETE


backend/tests/fixtures/halls.sql
```diff
@@ -0,0 +1,22 @@
...
+    'zaal-b',
+    '20000000-0000-0000-0000-000000000002'
+),
+(
+    '30000000-0000-0000-0000-000000000003',
+    3003,
+    'Studio Hal',
+    'studio-hal',
+    '20000000-0000-0000-0000-000000000003'
+);
```
hanneskt, line 29: suggestion: assign at least 2 halls to one of the spaces, and have a space without any halls

In the fixtures for m-1 relations, I try to make sure there is one of each case: 0-1, 1-1 and >=2-1. This often catches weird edge cases.

I would add this now, but this will probably become more relevant when we return joined objects (a space returns a list of it's halls as well, for example) where we can use this to test the different cases in the object above (a space with 2, 1 and 0 halls). We can test for it whenever we add that feature

backend/tests/fixtures/spaces.sql
```diff
@@ -0,0 +1,19 @@
...
+    2002,
+    'Kleine Zaal',
+    '10000000-0000-0000-0000-000000000001'
+),
+(
+    '20000000-0000-0000-0000-000000000003',
+    2003,
+    'Studio',
+    '10000000-0000-0000-0000-000000000002'
+);
```
hanneskt, line 19: praise: you did here what I talked about, each case in the m-1 relation is covered!

hanneskt (changes requested):
Basically good. Just have the fixture cover the different cases.

The tests are well written otherwise!

backend/tests/fixtures/halls.sql
jarnex47: alright, will do!

hanneskt (approved):
Alright great!

Kaya-Sem (approved):

Issue #111, opened 2026-03-15
feat: backend for superadmin


Thoosje:
**Is your feature request related to a problem? Please describe.**
Currently nobody can create admins. A superadmin needs to be able to create and edit admins. (Possibly also add password reset)

**Describe the solution you'd like**
Add enum in the database that keeps track of what type of user the user is: `admin` or `superadmin`. Incase we want to make this
setup future proof we can also add the type `user` which might be used for normal users if needed.

The extractor needs to keep track of this type. It should be easy to make routes that can only be accessed by a `superadmin`.
The `superadmin` has access to all resources that normal `admins` have access to.

**Describe alternatives you've considered**
Instead of a type simply add a bool: `is_super_admin`

**Additional context or Mockups**
The `superadmin` will possibly need extra page in the UI to manager all the admins and other settings only a superadmin can handle.

Kaya-Sem:
Can this issue be closed, as #120 is merged?

Issue #110, opened 2026-03-15
feat: protect CRUD routes


Thoosje:
**Is your feature request related to a problem? Please describe.**
Currently CRUD enpoints (`delete`, `patch` and `create`) are not protected by the authentication layer. This is needed to prevent
normal users editing the data. Only admin users should have access to these routes.

**Describe the solution you'd like**
Add the `auth` extractor to the endpoints and edit the swagger documentation. A small example is already implemented:
https://github.com/SELab-2/viernulvier-6/blob/main/backend/src/handlers/admin.rs

PR #120, opened 2026-03-18
feat(admin): admin and CRUD protection 

Thoosje:
# Superadmin and CRUD protection

**Description**
This PR introduces the `Editor` role which will take over the old `Admin` role, and provides an interface for admins (comparable to actual super admins) to create and manage other editor users. All the CRUD endpoints are protected (except GET) so only editors (and above) can edit data.

**Key Changes:**
* **Backend:** 
    * Implemented `RequireAdmin` and `RequireEditor` extractors to handle Authorization.
    * Applied `RequireEditor` to all CRUD endpoints (POST, PUT, PATCH, DELETE) across the API (e.g., Productions, Locations, Spaces, Halls) to ensure only authorized personnel can modify data.
    * Secured `/editor/me` to ensure only editors and admins can access their administrative profile.
* **Frontend:** * Implemented the `UserRole` enum and updated the `mapUser` function to normalize backend roles.
    * Integrated a localized `CreateAdminForm` and protected the Admin dashboard UI using the new role-based logic.
* **Fixes & Refactors:** 
    * Improved the `InputField` component to support labels and `useId` for better accessibility.
    * Standardized API calls using a central Axios instance with automated JWT refresh interceptors.

**Conclusion**
- Old `Admin` is now `Editor`. This naming has changed thru out the frontend and API - **BREAKING CHANGE**
- The new `Admin` can actually modify the `Editors`

**Related Issue**
Implements #111 
Implements #110

**Type of change**
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that causes existing functionality to change)
- [x] ♻Refactor / Chore / Documentation

**How Has This Been Tested?**
- [x] Local manual testing
- [x] Tested in the staging environment

**Developer Checklist:**
- [x] I have updated the documentation (if applicable).

image


Kaya-Sem (commented):
suggestion(role-naming): `admin` / `superadmin` are ambiguous and do not clearly reflect permission levels

proposed change:

* `superadmin` → `admin`
* `admin` → `editor`

rationale: “admin” is typically understood as the highest privilege level, while “editor” better communicates content-focused access. This should make roles more intuitive for CMS/admin dashboard users.

Thoosje:
> suggestion(role-naming): `admin` / `superadmin` are ambiguous and do not clearly reflect permission levels
> 
> proposed change:
> 
> * `superadmin` → `admin`
> * `admin` → `editor`
> 
> rationale: “admin” is typically understood as the highest privilege level, while “editor” better communicates content-focused access. This should make roles more intuitive for CMS/admin dashboard users.



> suggestion(role-naming): `admin` / `superadmin` are ambiguous and do not clearly reflect permission levels
> 
> proposed change:
> 
> * `superadmin` → `admin`
> * `admin` → `editor`
> 
> rationale: “admin” is typically understood as the highest privilege level, while “editor” better communicates content-focused access. This should make roles more intuitive for CMS/admin dashboard users.

I've done this, checked everything, and added the tests. When reviewing, please make sure I haven't forgotten to change anything, as small errors can easily be overlooked.

Kaya-Sem (approved):

backend/migrations/20260227093503_auth.up.sql
```diff
@@ -1,7 +1,10 @@
-CREATE TABLE users ( -- Call this users or admin_users?
+CREATE TYPE user_role AS ENUM ('admin', 'editor', 'user');
+
+CREATE TABLE users (
    id UUID PRIMARY KEY DEFAULT uuidv7(),
    username TEXT NOT NULL,
    email TEXT NOT NULL UNIQUE,
    password_hash TEXT NOT NULL,
+   role user_role NOT NULL DEFAULT 'user',
```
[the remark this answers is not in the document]
hanneskt: This is valid, especially for local dev. You can get away with it, but it requires some special actions (people drop their dev databases, prod database gets dropped) and it's easier to just but it in a separate migration.

backend/src/handlers/location.rs
```diff
@@ -46,10 +53,15 @@ pub async fn get_one(db: Database, Path(id): Path<Uuid>) -> JsonResponse<Locatio
...
-        (status = 201, description = "Created", body = LocationPayload)
+        (status = 201, description = "Created", body = LocationPayload),
+        (status = 401, description = "Unauthorized", body = ErrorResponse)
+    ),
+    security(
+        ("cookie_auth" = [])
     )
 )]
 pub async fn post(
+    _admin: RequireEditor,
```
hanneskt: issue: this works, but is very easy to miss when adding a new endpoint.

What I like to is to put protected routes in a different router, and protect the whole router using a layer. See [this implementation](https://github.com/ZeusWPI/ZPI/blob/653549d5c1819948145689059f0b21d1c0ba22ed/src/lib.rs#L99-L112).
I'm not fully certain that is the best way, though, so other suggestions are also welcome.

You can even just use your extractor for this:
```rust
.route_layer(from_extractor::<AuthUser>())
```

backend/src/extractors/auth.rs
```diff
@@ -34,10 +37,46 @@ impl FromRequestParts<AppState> for AuthUser {
...
+            return Err(AppError::Unauthorized);
+        }
+
+        Ok(RequireAdmin(user))
+    }
+}
+
+pub struct RequireEditor(pub AuthUser);
+
+impl FromRequestParts<AppState> for RequireEditor {
```
hanneskt: suggestion: rename these to `AdminUser` and `EditorUser`

(or just `Admin` and `Editor`)
For protections you should use layers, and use the extractors if you actually need that object. Having one for the editor and the admin is good, so you don't have to check permissions in every endpoint, and you're using the typesystem to make verify.

In short: these should be named for using the data in them, not for protecting routes.

See https://docs.rs/axum/latest/axum/middleware/ for more info about middleware and layers

backend/tests/location.rs
hanneskt, line 1: suggestion: add a check to see if the route is actually protected

Keep the tests the same, but check if it rejects an unauthenticated router first, then login and continue the tests like it has now.
Do this on all routes protected routes? This is what we have in[ another project](https://github.com/ZeusWPI/ZPI/blob/653549d5c1819948145689059f0b21d1c0ba22ed/tests/achievement.rs#L14-L32). But you can see if you want to do it like that or not.

A function on the router .login() or something that makes an account, logs in and saves the cookie for subsequent requests might be even handier.

backend/tests/common/user.rs
```diff
@@ -0,0 +1,69 @@
...
+    }).await.unwrap();
+
+    let exp = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_secs()
+        + (config.access_token_expiry_minutes as u64 * 60);
+
+    let claims = Claims {
+        sub: user.id,
```
hanneskt, line 55: suggestion: maybe add/use this to/in the router?

I did something similar in [another project](https://github.com/ZeusWPI/ZPI/blob/653549d5c1819948145689059f0b21d1c0ba22ed/tests/common/router.rs#L52-L60). It keeps the builder pattern of the TestRouter. The rest of that file also contains with admin users.

I think it decreases the boilerplate you need in the tests.

hanneskt (changes requested):
Good, but I think the guarding can be less error-prone

backend/src/handlers/admin.rs
```diff
@@ -1,34 +1,103 @@
...
     ),
     security(
         ("cookie_auth" = [])
     )
 )]
-pub async fn admin(
-    auth: AuthUser,
+pub async fn editor_me(
+    EditorUser(editor): EditorUser,
     _state: State<AppState>,
```
hanneskt: suggestion: This state can be removed

backend/migrations/20260227093503_auth.down.sql
```diff
@@ -1 +1 @@
-DROP TABLE IF EXISTS users; -- Add down migration script here
+DROP TABLE IF EXISTS users;
```
hanneskt:
```suggestion
DROP TABLE users;
```

backend/src/extractors/auth.rs
```diff
@@ -34,10 +37,46 @@ impl FromRequestParts<AppState> for AuthUser {
             id: token_data.claims.sub,
             email: token_data.claims.email,
             session_id: token_data.claims.sid,
+            role: token_data.claims.role,
         })
     }
 }
 
+#[allow(dead_code)] // the Authuser isn't actually being used for now but the guard is.
```
hanneskt, line 45: question: is this still needed?

backend/tests/common/router.rs
```diff
@@ -22,15 +26,38 @@ impl TestRouter {
...
+    }
+
+    pub async fn as_admin(db: PgPool) -> Self {
+        Self::new(db).login("admin@test.com", UserRole::Admin).await
+    }
+
+    pub async fn as_user(db: PgPool) -> Self {
+        Self::new(db).login("user@test.com", UserRole::User).await
+    }
+
```
hanneskt, line 60: praise: really clean now!

hanneskt (approved):
Some small tweaks, but the backend part looks good otherwise!

PR #109, opened 2026-03-15
feat(cms): add basic CMS frontend

Kaya-Sem:
Partially implements #19 

Add a cms page behind `[locale]/cms`. 
Features tabs for the different systems. There are tables for each entity, with support for subtables (eg. productions and their events). The different tabs and entity tables are routes.

Rows have the UI to edit (API not yet integrated)

Currently has login guard commented out. The current focused tab is in the URL params.

https://sel2-6.ugent.be/pr-109/nl/cms

PR #142, opened 2026-03-22
feat(cms): make fields for entity table rows editable

Kaya-Sem:
Wire CMS edit sheet to backend + seed admin user

  - Backend: fixed CORS to allow PUT/DELETE; 
- added computed metadata_status field to ProductionPayload (derived server-side. "complete" when all key bilingual content fields are present, "partial" otherwise)
 - added dev admin seed migration
- Frontend: updated Production type to mirror the full bilingual backend payload (title_nl/title_en, artist_nl/artist_en, etc.)
- added query hooks for productions, locations, halls and spaces; added mutation hooks (useUpdateProduction, useUpdateLocation, useUpdateHall) that PUT to the existing backend endpoints; wired onSave in both ProductionsTable and  VenuesTable.

 halls are joined to venues client-side via the space relationship atm.
-  fixed NEXT_PUBLIC_API_URL to point at localhost:3001/api .  might be fixed in another ongoing PR

>[!WARNING]
> contains tagging features from #104. Merge tagging first. Also #109


Kaya-Sem:
@Thoosje Seems this already works in main due to some work you did, can you confirm, so this can just be closed?

Thoosje:
> @Thoosje Seems this already works in main due to some work you did, can you confirm, so this can just be closed?

Pretty sure all of your comments should be implemented, except this one I need to check:
- added computed metadata_status field to ProductionPayload (derived server-side. "complete" when all key bilingual content fields are present, "partial" otherwise)

PR #152, opened 2026-03-23
fix(traefik): Fixed previews not working

Thoosje:
This PR adds the priority back as currently preview builds are being taken over by the production version. 

For example: https://sel2-6.ugent.be/pr-150/api is redirecting to https://sel2-6.ugent.be/nl/pr-150/api


brentvent (approved):
Nice, thanks

tst-038 (approved):

PR #155, opened 2026-03-23
fix: make API_KEY_404 optional to prevent crash on startup

### LINKED ISSUE #89: [BACKEND] API Should Not Crash Without API_KEY_404
## Problem

Currently, if the `API_KEY_404` environment variable is missing, the API crashes on startup. This breaks the system even if the **data importer** is not needed.

The importer is important, but the API itself should remain functional without it.

## Proposed Solution

- Make the `API_KEY_404` optional
- Only enable the importer if the key is present
- Log a warning if the key is missing instead of crashing

tst-038:
**Description**
- Change api_key_404 from String to Option<String>
- Only start API importer when key is present
- Log warning when API_KEY_404 is missing

Fixes: API crashes on startup when API_KEY_404 env var is missing

**Related Issue**
Fixes #89 

**How Has This Been Tested?**
- [x] Local manual testing

**Developer Checklist:**


Thoosje (approved):

backend/src/config.rs
```diff
@@ -23,9 +23,14 @@ impl AppConfig {
             .map(|s| s.trim().to_string())
             .collect();
 
+        let api_key_404 = env::var("API_KEY_404").ok();
+        if api_key_404.is_none() {
+            warn!("API_KEY_404 not set, API importer will be disabled");
+        }
+
```
hanneskt, line 30: suggestion: log this warning where you start the api importer

If the logic ever changes, the tracing is disconnected from the function and could be forgotten.

hanneskt (changes requested):

backend/src/config.rs
tst-038: Valid, but this was done intentionally. This way all warnings/errors about env variables are grouped together at startup, making them immediately visible. Checking these later won’t change much and might result in unnoticed warnings. By logging these at startup, it’s easier to catch configuration issues early.

hanneskt: you can definitely warn here warn!("API_KEY_404 not set") but i would warn the "API importer will be disabled" where the logic is to disable.

Ultimately, it doesn't matter much, so I'll approve and you can choose if you want to make that change or not.

hanneskt (approved):

Issue #137, opened 2026-03-22
Set up test coverage


jerjoen:
## Scope
- Set up cargo-llvm-cov
- Add README instructions
- Generate initial report

PR #161, opened 2026-03-24
docs(coverage): Test coverage setup

jerjoen:
Update README for coverage setup

implements #137

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -52,6 +52,25 @@ cargo test --all
 
 This requires a running database.
 
+## Test coverage
+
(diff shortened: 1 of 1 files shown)


hanneskt (approved):
Nice, works!

tst-038 (approved):

tst-038:
add cargo-llvm-cov as dev dependency

hanneskt:
> add cargo-llvm-cov as dev dependency

Sadly not how that works, that would add it as a library, not a cli tool

PR #181, opened 2026-03-26
feat: filter spec

Kaya-Sem:
**Description**
introduces a draft spec for filtering. Places in /spec for lack of better place

**Related Issue**
Fixes #

**How Has This Been Tested?**
- [ ] Local manual testing

**Developer Checklist:**

### DIFF
--- a/specs/search-filter-api.md
+++ b/specs/search-filter-api.md
@@ -1,0 +1,334 @@
+# Search & Filter API Spec
+
+**Status:** Draft
+**Scope:** Backend implementation spec for entity querying — filterable list endpoints and unified search.
+
+---
+
+## 1. Overview
+
+The archive exposes two query surfaces:
+
+1. **Per-type filtered list endpoints**: used by the CMS and public browsing pages for a single entity type.
+2. **Unified search endpoint**: used by the public search page, returns interleaved results across entity types.
+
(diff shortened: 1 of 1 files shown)


Thoosje:
@hanneskt Please give your opinion on this when you have time

brentvent (approved):
Looks like a solid start. Some questions:
- the response for single-type endpoints look akin to the one of unified search (with the next-cursor) right?
- Are the relevance scores going to work properly for the unified search?

I would also include search on substrings inside an event's description etc, but maybe this is a case of 'toeters en bellen' and thus for milestone 3.

Kaya-Sem:
> Looks like a solid start. Some questions:
> 
>     * the response for single-type endpoints look akin to the one of unified search (with the next-cursor) right?
> 
>     * Are the relevance scores going to work properly for the unified search?
> 
> 
> I would also include search on substrings inside an event's description etc, but maybe this is a case of 'toeters en bellen' and thus for milestone 3.

To answer relevance scores: this depends on the implementation, really. I hope it will work!
For substrings in more descriptions, I think that would pollute the search results alot, so I am hesistant for that

hanneskt (approved):
Looks good. I've started implementation and we can course correct if anything turns out to not work well.

PR #184, opened 2026-03-27
feat(backend): add backend support for prices

### LINKED ISSUE #183: feat(backend): add more endpoints
We currently have the following endpoints:
- production
- event
- location
- space
- hall

Suggestion to add the following (feel free to add):
- price
- reeksen?

brentvent:
**Description**
Imported ```prices``` and ```price_ranks```. ```Prices``` are currently only made relevant inside ```events```,  and thus nested inside the JSON response for ```events``` (as per suggestion of @Thoosje and @Kaya-Sem). Dissenting opinions welcome: group 2 DO present them as seperate endpoints (https://sel2-2.ugent.be/api/docs/), but I also do not see why WE would do this.

Some more context: ```events``` contain a ```prices``` field which in turn contains ```event_price```, ```price``` and ```price_rank```. A bit confusing, but ```event_price``` contains the actual price in cents (their API provided a string for this, so I parsed it into cents, as parsing it into a float is risky), ```price``` is more like the ticket 'type' and ```price_rank``` contains the 'class' its in.

TODO:
- [x] import prices and nest them inside events
- [x] add test for prices (importer excluded)
- [x] add hooks to frontend

**Related Issue**
Fixes #183 

**How Has This Been Tested?**
- [x] Local manual testing

**Developer Checklist:**


Thoosje:
@brentvent do not forget to add the hooks needed in the frontend, you can take a look at the other hooks.

Kaya-Sem:
> @brentvent do not forget to add the hooks needed in the frontend, you can take a look at the other hooks.

suggestion: preferably prices would be packaged with the entity it comes from, and no extra endpoinds or DTOs imo

Thoosje (approved):

Thoosje:
Make sure all the frontend hooks and tests continue to be working as this might make it a little bit different (and make sure to use the auto generated types). Looks good, also @hanneskt maybe you should review aswel

backend/api/src/helper.rs
```diff
@@ -48,4 +73,21 @@ mod tests {
...
+        assert_eq!(parse_amount_cents("10.00"), Some(1000));
+        assert_eq!(parse_amount_cents("10.50"), Some(1050));
+        assert_eq!(parse_amount_cents("10.75"), Some(1075));
+        assert_eq!(parse_amount_cents("-1.25"), Some(-125));
+    }
+
+    #[test]
+    fn test_parse_amount_cents_invalid_inputs() {
+        assert_eq!(parse_amount_cents("10"), None);
+        assert_eq!(parse_amount_cents("10.5"), None);
```
hanneskt, line 89: question: are these not valid?

They look like valid price representations? Does the api alway return something of the shape `1.11`?

backend/src/dto/event.rs
```diff
@@ -23,13 +25,11 @@ impl EventPayload {
...
-            .all(limit + 1, id_cursor)
-            .await?
-            .into_iter()
-            .map(Self::from)
-            .collect();
+        let events = db.events().all(limit + 1, id_cursor).await?;
+        let mut data = Vec::with_capacity(events.len());
+        for event in events {
+            data.push(Self::from_model(db, event).await?);
+        }
```
hanneskt, line 33: question: why was this changed from a `map` to a explicit for loop?

Map also takes a closure if you need that

backend/src/dto/event.rs
```diff
@@ -43,76 +43,227 @@ impl EventPayload {
...
-            .by_production(id)
-            .await?
-            .into_iter()
-            .map(Self::from)
-            .collect())
+        let events = db.events().by_production(id).await?;
+        let mut payloads = Vec::with_capacity(events.len());
+        for event in events {
+            payloads.push(Self::from_model(db, event).await?);
+        }
```
hanneskt, line 56: question: why was this changed to an explicit loop?

hanneskt (commented):
Looks good! I have some question to make things clearer though.

backend/api/src/helper.rs
brentvent: The API does indeed always return prices of shape ```X.XX```, but in principle something like ```10.5``` is not wrong either.

backend/src/dto/event.rs
brentvent: We need a DB lookup for ```prices```, and thus this part becomes ```async```. Apparently this does not work with ```map```.

backend/src/dto/event.rs
brentvent: same reason as above

backend/src/dto/event.rs
```diff
@@ -43,76 +43,227 @@ impl EventPayload {
...
+        Ok(payloads)
     }
 
     pub async fn update(self, db: &Database) -> Result<Self, AppError> {
-        Ok(db.events().update(self.into()).await?.into())
+        let (event, prices) = self.into_parts();
+        let updated = db.events().update(event).await?;
+        EventPricePayload::sync_for_event(db, updated.id, prices).await?;
```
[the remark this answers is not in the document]
brentvent: This is how it's done for every endpoint.

hanneskt (approved):

PR #189, opened 2026-03-28
feat(articles): create endpoints and basic editor in CMS

Kaya-Sem:
**Description**

Implements the full article feature end-to-end: database schema, backend CRUD endpoints, and a CMS editor with rich-text support.

**Backend**
- `articles` migration (3 migration steps, landing on a single `title` + single `content` column)
- `Article` model, DTO, repository, and handler in the database/backend crates
- New routes wired into the Axum router
- Updated `.sqlx` offline query cache

**Frontend**
- New route group `cms/(editor)/articles/[id]/edit/` with a full-page article editor
- `TiptapEditor` component with a custom toolbar (`editor-toolbar.tsx`)
- `ArticleMetadataPanel` for title, status, and related fields
- `StatusBadge` component
- Articles table with columns under `cms/tables/articles/`
- `useArticles` / `useArtists` TanStack Query hooks
- `article.mapper` and updated `artist.mapper`
- i18n strings added in `en.json` / `nl.json`

> The large diff is mostly `backend/.sqlx` generated query files

image

image


backend/database/src/repos/article.rs
```diff
@@ -0,0 +1,245 @@
...
+                "event" => Some("articles_about_events"),
+                _ => None,
+            };
+            let id_col = match entity_type.as_str() {
+                "production" => Some("production_id"),
+                "artist" => Some("artist_id"),
+                "location" => Some("location_id"),
+                "event" => Some("event_id"),
+                _ => None,
+            };
```
hanneskt: suggestion: do this with an enum

The enum can be used in the api as well to deserialize the json. You can then functions on the enum to get the correct tables and columns.

hanneskt (changes requested):
Looks okay, but needs tests!

backend/database/src/repos/article.rs
```diff
@@ -0,0 +1,266 @@
...
+            );
+            builder.push_bind(slug);
+            builder.push(")");
+        }
+
+        if let (Some(entity_id), Some(entity_type)) = (related_entity_id, related_entity_type) {
+            if let (Some(table), Some(col)) = (
+                entity_type.article_join_table(),
+                entity_type.article_id_column(),
+            ) {
```
hanneskt: suggestion: write this as a let-&& chain

```rust
if let Some(...) = ...
	&& let Some(...) = ... {

}
```
This is a bit cleaner than a double nest

If you configure your rust-analyzer to use `cargo clippy` it would have told you :p

Image

backend/src/utils.rs
```diff
@@ -0,0 +1,10 @@
+pub fn slugify(s: &str) -> String {
+    s.to_lowercase()
+        .chars()
+        .map(|c| if c.is_alphanumeric() { c } else { '-' })
+        .collect::<String>()
+        .split('-')
+        .filter(|p| !p.is_empty())
+        .collect::<Vec<_>>()
+        .join("-")
+}
```
hanneskt: suggestion: use [slug](https://lib.rs/crates/slug)

We use this already for other slugs

backend/src/dto/article.rs
```diff
@@ -0,0 +1,237 @@
...
+        } else {
+            let mut n = 2u32;
+            loop {
+                let candidate = format!("{base_slug}-{n}");
+                if !db.articles().slug_exists(&candidate).await? {
+                    break candidate;
+                }
+                n += 1;
+            }
+        };
```
hanneskt: suggestion: can this be improved by having the database query return how many there are already?

This can probably be done with some kind of regex match maybe? It would be a lot cleaner than to make a query for every single check.

hanneskt (approved):
Good after those comments

PR #201, opened 2026-03-30
fix(hooks): Add Pagination Support to API Hooks and Types

Thoosje:
### Short Description

This PR adds comprehensive pagination support to all TanStack Query hooks, updates API types to use operation-specific response types, implements infinite scroll on the search page, and removes the unused numbered pagination component. All list endpoints now support cursor-based pagination with proper TypeScript typing.

---

#### 1. API Type System Overhaul

**Added utility types in `api.types.ts`:**

- `SuccessResponse<T>` - Extracts 200/201 response content from operations
- `PaginatedListResponse<T>` - Extracts paginated list responses
- `PaginationParams` - Interface for cursor/limit parameters
- `PaginatedResult<T>` - Standard paginated result structure with `data` and `nextCursor`

**Updated entity type files to use operation-specific types:**

- `event.api.types.ts` - `GetEventByIdResponse`, `CreateEventResponse`, `UpdateEventResponse`, `GetAllEventsResponse`
- `production.api.types.ts` - `GetProductionByIdResponse`, `CreateProductionResponse`, etc.
- `hall.api.types.ts`, `location.api.types.ts`, `space.api.types.ts` - Same pattern
- `auth.api.types.ts`, `taxonomy.api.types.ts` - Updated to use `SuccessResponse<>` utility

#### 2. Pagination Support in Hooks

**All list hooks now support cursor-based pagination:**

```typescript
// New hook signature with optional pagination
useGetEvents(options?: { enabled?: boolean; pagination?: PaginationParams })
useGetProductions(options?: { enabled?: boolean; pagination?: PaginationParams })
useGetHalls(options?: { enabled?: boolean; pagination?: PaginationParams })
useGetLocations(options?: { enabled?: boolean; pagination?: PaginationParams })
useGetSpaces(options?: { enabled?: boolean; pagination?: PaginationParams })
```

**Query keys updated for pagination:**

- `queryKeys.events.all(pagination?)` - Returns `["events"]` or `["events", {cursor, limit}]`
- Different pagination params create separate cache entries
- Mutations invalidate all paginated queries for that entity

#### 3. Mapper Functions

**Added `mapPaginatedXxxResult` functions to all mappers:**

- `mapPaginatedEventsResult()` - Returns `PaginatedResult<Event>`
- `mapPaginatedProductionsResult()` - Returns `PaginatedResult<Production>`
- `mapPaginatedHallsResult()` - Returns `PaginatedResult<Hall>`
- `mapPaginatedLocationsResult()` - Returns `PaginatedResult<Location>`
- `mapPaginatedSpacesResult()` - Returns `PaginatedResult<Space>`

These extract both the data array and nextCursor from API responses.

#### 4. Component Updates

**Search page (`/search/page.tsx`):**

- Implemented infinite scroll using Intersection Observer
- Removed numbered pagination (component deleted)
- Accumulates loaded productions to prevent duplicates
- Shows loading spinner when fetching more
- Disables infinite scroll during client-side search filtering
- Resets to first page when clearing search

**Updated all consuming components:**

- `page.tsx` (homepage) - Uses `productionsResult?.data`
- `search/page.tsx` - Uses cursor-based infinite scroll
- `locations-table.tsx` - Updated for paginated results
- `productions-table.tsx` - Updated for paginated results

#### 5. MSW Handlers Updated

**Test mocks now return paginated responses:**

- `events.handlers.ts` - Returns 2 pages with cursor navigation
- `productions.handlers.ts`, `halls.handlers.ts`, etc. - Return `{ data: [...], next_cursor: ... }`

#### 6. Removed Unused Code

**Deleted components:**

- `/components/searchpage/pagination/Pagination.tsx`
- `/components/searchpage/pagination/index.ts`

#### 7. Test Coverage

**New pagination tests added:**

- `useEvents.test.tsx` - Tests first page, next page, limit param, combined params, cache isolation
- `useLocations.test.tsx`, `useProductions.test.tsx`, etc. - Basic paginated result tests
- `query-keys.test.ts` - Tests pagination key generation

**Updated existing tests:**

- All hook tests now use `queryKeys.xxx.all()` (function call syntax)
- Resource mutation tests updated for new query key format
- MSW handlers return proper paginated structure

---

### Breaking Changes

1. **Hook return types changed** - All list hooks now return `PaginatedResult<T>` instead of `T[]`
    - Before: `const { data: events } = useGetEvents()` // Event[]
    - After: `const { data: result } = useGetEvents()` // PaginatedResult<Event>
    - Access data: `result?.data ?? []`
    - Access next cursor: `result?.nextCursor`

2. **Query keys are now functions** - Must call `.all()` instead of using `.all`
    - Before: `queryKeys.events.all`
    - After: `queryKeys.events.all()` or `queryKeys.events.all(paginationParams)`

3. **Pagination component removed** - If any other pages used it, they need to migrate to infinite scroll or a new pagination approach.

---

### Migration Guide for Developers

```typescript
// Before
const { data: productions } = useGetProductions();
const filtered = productions?.filter(...) ?? [];

// After
const { data: result } = useGetProductions();
const productions = result?.data ?? [];
const hasMore = result?.nextCursor !== null;

// Before
queryClient.invalidateQueries({ queryKey: queryKeys.productions.all });

// After
queryClient.invalidateQueries({ queryKey: queryKeys.productions.all() });
```

---

### Testing

All 77 tests pass:

- Unit tests for mappers
- Integration tests for hooks with pagination
- Component behavior tests
- Query key tests

Run tests: `npm test`
Check types: `npm run typecheck`


frontend/src/hooks/api/useProductions.ts
```diff
@@ -47,14 +58,14 @@ export const useCreateProduction = () => {
...
-            const { data } = await api.post<ProductionResponse>(
+            const { data } = await api.post<CreateProductionResponse>(
                 "/productions",
                 mapCreateProductionInput(payload)
             );
             return mapProduction(data);
         },
         onSuccess: () => {
-            queryClient.invalidateQueries({ queryKey: queryKeys.productions.all });
+            queryClient.invalidateQueries({ queryKey: ["productions"] });
```
tst-038: Shouldn't this be `queryKeys.productions.all()`? Otherwise paginated querries wouldn't get invalidated.

frontend/src/hooks/api/useProductions.ts
```diff
@@ -64,14 +75,14 @@ export const useUpdateProduction = () => {
...
-            const { data } = await api.put<ProductionResponse>(
+            const { data } = await api.put<UpdateProductionResponse>(
                 "/productions",
                 mapUpdateProductionInput(payload)
             );
             return mapProduction(data);
         },
         onSuccess: (production) => {
-            queryClient.invalidateQueries({ queryKey: queryKeys.productions.all });
+            queryClient.invalidateQueries({ queryKey: ["productions"] });
```
tst-038: same thing as [here](https://github.com/SELab-2/viernulvier-6/pull/201/changes#r3022093818)

frontend/src/hooks/api/useProductions.ts
```diff
@@ -86,7 +97,7 @@ export const useDeleteProduction = () => {
             return id;
         },
         onSuccess: (id) => {
-            queryClient.invalidateQueries({ queryKey: queryKeys.productions.all });
+            queryClient.invalidateQueries({ queryKey: ["productions"] });
```
tst-038: same thing as [here](https://github.com/SELab-2/viernulvier-6/pull/201/changes#r3022093818)

frontend/src/app/[locale]/(app)/search/page.tsx
```diff
@@ -17,40 +19,109 @@ import { SearchHero } from "@/components/searchpage/search-hero";
...
+            const timeoutId = setTimeout(() => {
+                setAllProductions((prev) => {
+                    const newIds = new Set(productionsData.map((p) => p.id));
+                    const filtered = prev.filter((p) => !newIds.has(p.id));
+                    return [...filtered, ...productionsData];
+                });
+            }, 0);
+            return () => clearTimeout(timeoutId);
+        }
+    }, [productionsData, currentPageIndex, cursorHistory.length]);
```
Comment (tst-038) [inline frontend/src/app/[locale]/(app)/search/page.tsx]: quite hacky, can cause missed pages, runs after render
```typescript
Gebruik een ref om bij te houden welke pagina's al verwerkt zijn:
  const processedCursors = useRef<Set<string>>(new Set());

  useEffect(() => {
      const cursorKey = currentCursor ?? 'initial';

      // Alleen toevoegen als deze cursor nog niet verwerkt is
      if (productionsData.length > 0 && !processedCursors.current.has(cursorKey)) {
          processedCursors.current.add(cursorKey);

          setAllProductions((prev) => {
              const newIds = new Set(productionsData.map((p) => p.id));
              const filtered = prev.filter((p) => !newIds.has(p.id));
              return [...filtered, ...productionsData];
          });
      }
  }, [productionsData, currentCursor]);

  // Reset bij zoekopdracht/filter reset
  const handleReset = useCallback(() => {
      setCursorHistory([null]);
      setCurrentPageIndex(0);
      setAllProductions([]);
      processedCursors.current.clear();  // ← Ook de ref resetten!
  }, []);
```

tst-038 (changes requested):
Use query keys more consistently, sometimes they are being used sometimes they aren't

jerjoen (approved):
Ran `npm test` and `npm run typecheck` locally and all 77 tests pass. Tests cover the key pagination scenarios well. Minor thing: didn't see explicit error handling tests (e.g., API 500), not blocking, but worth considering?

Hook signatures, mapper functions and MSW handlers all look consistent.
Branch is out-of-date with main --> rebase before merging.

frontend/src/app/[locale]/(app)/search/page.tsx
```diff
@@ -17,40 +19,116 @@ import { SearchHero } from "@/components/searchpage/search-hero";
...
+    const currentCursor = cursorHistory[currentPageIndex];
+
+    const {
+        data: productionsResult,
+        isLoading: productionsLoading,
+        isFetching,
+    } = useGetProductions({
+        pagination: currentCursor ? { cursor: currentCursor } : undefined,
+    });
+    const { data: eventsResult, isLoading: eventsLoading } = useGetEvents();
```
Comment (Kaya-Sem) [inline frontend/src/app/[locale]/(app)/search/page.tsx]: are we fetching ALL events here, to group them client-side by production ID (lines 82-89)? 
Let's use `useGetEventsByProduction(productionId)` to fetch only the events for productions. There will always be many more events that productions, so negates pagination for productions.

frontend/src/app/[locale]/(app)/search/page.tsx
```diff
@@ -17,40 +19,116 @@ import { SearchHero } from "@/components/searchpage/search-hero";
...
     const searchParams = useSearchParams();
+    const loadMoreRef = useRef<HTMLDivElement>(null);
 
     const [searchQuery, setSearchQuery] = useState(() => searchParams.get("q") ?? "");
 
-    const { data: productions, isLoading: productionsLoading } = useGetProductions();
-    const { data: events, isLoading: eventsLoading } = useGetEvents();
-    const { data: locations, isLoading: locationsLoading } = useGetLocations();
+    // Cursor-based pagination state
+    const [cursorHistory, setCursorHistory] = useState<(string | null)[]>([null]);
```
Comment (Kaya-Sem) [inline frontend/src/app/[locale]/(app)/search/page.tsx:30]: This seems like reimplementing what `useInfiniteQuery` does, is this for a good reason?

frontend/src/app/[locale]/(app)/search/page.tsx
```diff
@@ -17,40 +19,116 @@ import { SearchHero } from "@/components/searchpage/search-hero";
...
+    // Accumulate productions when new page loads
+    useEffect(() => {
+        const cursorKey = currentCursor ?? "initial";
+
+        // Only add if this cursor hasn't been processed yet
+        if (productionsData.length > 0 && !processedCursors.current.has(cursorKey)) {
+            processedCursors.current.add(cursorKey);
+
+            // Defer state update to avoid cascading renders
+            queueMicrotask(() => {
```
Comment (Kaya-Sem) [inline frontend/src/app/[locale]/(app)/search/page.tsx]: Was there a specific render cascade or warning you were trying to fix? or can this wrapper be dropped? React 18 batches state updates in effects automatically as far as I know.

frontend/src/app/[locale]/(app)/search/page.tsx
```diff
@@ -17,40 +19,116 @@ import { SearchHero } from "@/components/searchpage/search-hero";
...
+    const handleReset = useCallback(() => {
+        setCursorHistory([null]);
+        setCurrentPageIndex(0);
+        setAllProductions([]);
+        processedCursors.current.clear(); // Also reset the ref!
+    }, []);
+
+    // Infinite scroll with Intersection Observer
+    useEffect(() => {
+        if (searchQuery) return; // Disable infinite scroll when searching
```
Comment (Kaya-Sem) [inline frontend/src/app/[locale]/(app)/search/page.tsx]: When a user types a search, he gets results only from already-loaded pages, and has no way to load more?

frontend/src/app/[locale]/(app)/search/page.tsx
```diff
@@ -64,30 +142,28 @@ export default function SearchPage() {
...
 
-    const allProductions = productions ?? [];
-    const totalPages = Math.max(1, Math.ceil(allProductions.length / ITEMS_PER_PAGE));
-    const pagedProductions = allProductions.slice(
-        (currentPage - 1) * ITEMS_PER_PAGE,
-        currentPage * ITEMS_PER_PAGE
-    );
-
+    // Use accumulated productions for display
     const filteredProductions = searchQuery
```
Comment (Kaya-Sem) [inline frontend/src/app/[locale]/(app)/search/page.tsx]: This filters 'allProductions`, which only contains productions loaded so far. A production on a page that hasn't been loaded will never appear...

frontend/src/app/[locale]/(app)/search/page.tsx
```diff
@@ -17,40 +19,116 @@ import { SearchHero } from "@/components/searchpage/search-hero";
...
+    }, [eventsData]);
+
+    const loadMore = useCallback(() => {
+        if (nextCursor && !isFetching) {
+            setCursorHistory((prev) => [...prev, nextCursor]);
+            setCurrentPageIndex((prev) => prev + 1);
+        }
+    }, [nextCursor, isFetching]);
+
+    const handleReset = useCallback(() => {
```
Comment (Kaya-Sem) [inline frontend/src/app/[locale]/(app)/search/page.tsx]: When the user clears searchbar, `processedCursors` is cleared too, so the effect will re accumulate when the query refires. Can it happen that when the query for cursor `null` (page 1) is still cached in tanstack query? So productionsdata updates instantly, the effect runs, and re-addss page 1 data, but pages 2+ are lost. I don't think we want to reset scroll position on search clear and lose their position?

frontend/src/mappers/event.mapper.ts
Kaya-Sem: these should be set in the backend! also `mapUpdateEventInput` reuses it, so update requests also get fresh created_at from frontend, overwriting the original creation timestamp.

frontend/src/app/[locale]/(app)/search/page.tsx
Comment (Thoosje) [reply frontend/src/app/[locale]/(app)/search/page.tsx]:
```
ESLint:

Error: Calling setState synchronously within an effect can trigger cascading renders

Effects are intended to synchronize state between React and external systems such as manually updating the DOM, state management libraries, or other platform APIs. In general, the body of an effect should do one or both of the following:
* Update external systems with the latest state from React.
* Subscribe for updates from some external system, calling setState in a callback function when external state changes.

Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect).
```

I kept receiving this error so this was a hacky work around. Couldn't figure out how else to do it.

Kaya-Sem:
search/page.tsx: Let's hold off on client-side search for now since server-side search is coming. The auto-loading workaround adds complexity and still gives incomplete results until all pages are fetched.

Kaya-Sem:
event.mapper.ts: Update mapper is fixed, but `mapCreateEventInput` still sets `created_at` and `updated_at` on the frontend. These should be set by the backend, the frontend shouldn't be responsible for authoritative timestamps.

Kaya-Sem (approved):

jarnex47 (approved):
looks good, didnt know invalidateQueries() does prefix matching

PR #207, opened 2026-03-31
feat(backend): production endpoint allows searching

hanneskt:
**Description**
The production endpoint now allows searching

**How Has This Been Tested?**
- [x] Local manual testing

**Developer Checklist:**


Thoosje:
I think we should merge #201  before this and also adapt the hooks in this PR?

backend/src/handlers/helpers/pagination_query.rs
Kaya-Sem, line 1: this seems not used anywhere at the moment. Is this intending for other functionality?

backend/database/src/repos/production.rs
```diff
@@ -39,19 +43,112 @@ impl<'a> ProductionRepo<'a> {
...
+                    .push_bind(&search_q)
+                    .push(" <% full_search_text")
+                    .push(" GROUP BY production_id ");
+
+                // use the cursor if there is one
+                if let Some(cursor) = cursor
+                    && let Some(score) = cursor.score
+                {
+                    // HAVING score > cursor.score
+                    query.push(" HAVING MAX( ");
```
Kaya-Sem: Won't we have issues with precedence for the OR and HAVING binding?

Kaya-Sem:
How will filtering on tags and facets work? Is this in scope for this PR?

hanneskt:
> How will filtering on tags and facets work? Is this in scope for this PR?

This is for a next PR, I wanted to keep this one focussed

backend/database/src/repos/production.rs
hanneskt: I'm not exactly sure what you mean.

My query's HAVING is currently (as commented)
`HAVING score > cursor.score OR (score = cursor.score AND id < cursor.id)`
It works like it should right now. I could add safetybrackets though
`HAVING (score > cursor.score OR (score = cursor.score AND id < cursor.id))`

hanneskt: see https://www.postgresql.org/docs/18/sql-select.html maybe?

Kaya-Sem (approved):

Thoosje (approved):

jerjoen (approved):

backend/database/src/repos/production.rs
brentvent: shouldn't this be ```MIN```?

Thoosje: Good catch, I think so

Thoosje (changes requested):

PR #209, opened 2026-04-01
fix(auth): restore middleware.ts for auth redirects

tst-038:
**Description**
Restore middleware.ts convention for auth redirects. Next.js 16's proxy.ts has issues with cookie reading/execution order, causing login redirects to fail silently. Using middleware.ts (still supported) fixes this.

This issue seems to only take place in safari (tested in macOS Tahoe 26.2)

**Related Issue**
login redirect not working, remained on login page after login 200 Ok

**How Has This Been Tested?**
- [x] Local manual testing (Safari & chrome)

**Developer Checklist:**

### DIFF
--- a/frontend/middleware.ts
+++ b/frontend/middleware.ts
@@ -4,7 +4,7 @@ import { routing } from "@/i18n/routing";
 
 const intlMiddleware = createIntlMiddleware(routing);
 
-export function proxy(request: NextRequest) {
+export function middleware(request: NextRequest) {
     const token = request.cookies.get("access_token")?.value;
     const pathname = request.nextUrl.pathname;
 


frontend/middleware.ts
```diff
@@ -4,7 +4,7 @@ import { routing } from "@/i18n/routing";
 
 const intlMiddleware = createIntlMiddleware(routing);
 
-export function proxy(request: NextRequest) {
+export function middleware(request: NextRequest) {
```
Thoosje, line 7: Maybe add a comment that this method actually is deprecated: https://nextjs.org/docs/app/api-reference/file-conventions/proxy

(If this fixes safari we can use it for now)

Kaya-Sem (approved):

PR #211, opened 2026-04-01
feat(backend): improve importer error handling

### LINKED ISSUE #62: improve error handling throughout the api ingest crate

### LINKED ISSUE #85: [Backend] Valnet automatische archivering
Valnet voor gefaalde archivaties van viernulvier API.

Nieuwe velden, data, ontbrekende dingen opvangen en melden in CMS, en proberen de admins dit te resolven zonder code moeten aanpassen?

brentvent:
**Description**
Currently, when something goes wrong in the importer, it almost always just crashes. I added the following  behaviour:

  - conversion code such as ```to_create()``` (the one that coverts stuff like ```ApiProduction``` into ```ProductionCreate```) no longer panics on invalid API input. It now returns structured errors, which are
    handled by the importer loop.
  - in the import loop, item-level failures are handled per item instead of crashing the whole importer. Invalid relation
    references and invalid ```source_id``` values are explicitly detected; the affected item is skipped or partially imported depending
    on the relation.
  - importer errors and warnings are now explicit Rust enums instead of hardcoded strings, using the [thiserror crate](https://docs.rs/thiserror/latest/thiserror/).
  - the importer now supports partial progress instead of blocking on missing optional relations. Current behavior:

  | Entity | Relation | Required by importer? | If missing now | What happens later? |
  |---|---|---:|---|---|
  | space | location | Yes | space is skipped | it can be inserted/updated later if it is fetched again and the relation exists by then |
  | hall | space | No | hall is imported with space_id = None | a later import can fill in space_id |
  | event | production | Yes | event is skipped | it can be inserted/updated later if it is fetched again and the relation exists by then |
  | event | hall | No | event is imported with hall_id = None | a later import can fill in hall_id |

  - the importer now retries non-crashing fetch failures with exponential backoff (currently 3 retries starting at 250 ms).
  
  - importer item errors/warnings are now stored in a new ```import_errors``` DB table.
  
  - unresolved importer errors are exposed to an editor endpoint in the backend.
  
  - importer errors are now visible in the CMS in a read-only “Automatische import” tab.

## Next step
The CMS tab for the importer now mostly focusses on errors related to the items themselves. I'd make a seperate frontend for the importer itself, so the status can be checked, as wel as more serious importer-run errors such as network fails, fails with the viernulvier API,... But that is probably not a Milestone 2 issue.

**Related Issue**
Fixes #62 
Fixes #85 

**How Has This Been Tested?**
- [ ] Local manual testing

**Developer Checklist:**

### DIFF
--- a/backend/api/Cargo.toml
+++ b/backend/api/Cargo.toml
@@ -22,7 +22,7 @@ ai-normalization = ["dep:scraper"]
 [dependencies]
 database = { path = "../database" }
 tracing = { workspace = true }
-tokio = { workspace = true, features = ["fs", "io-util", "rt", "rt-multi-thread", "macros"] }
+tokio = { workspace = true, features = ["fs", "io-util", "rt", "rt-multi-thread", "macros", "time"] }
 tracing-subscriber = { version = "=0.3", default-features = false, features = ["env-filter", "fmt", "ansi"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/backend/api/src/error.rs
+++ b/backend/api/src/error.rs
@@ -1,5 +1,186 @@
+use database::{error::DatabaseError, models::import_error::ImportErrorCreate};
 use thiserror::Error;
+use uuid::Uuid;
 
+// ImportEntity, ImportField and ImportRelation are small enums that allow
+// for stuff like ImportEntity::Event, removing strings from the picture
+
+/// Speciefies what entity we are importing
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ImportEntity {
+    Event,
+    EventPrice,
+    Gallery,
+    Hall,
+    Location,
+    Media,
+    MediaVariant,
+    Price,
+    PriceRank,
+    Production,
+    Space,
+}
+
+impl ImportEntity {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Event => "event",
+            Self::EventPrice => "event_price",
+            Self::Gallery => "gallery",
+            Self::Hall => "hall",
+            Self::Location => "location",
+            Self::Media => "media",
+            Self::MediaVariant => "media_variant",
+            Self::Price => "price",
+            Self::PriceRank => "price_rank",
+            Self::Production => "production",
+            Self::Space => "space",
+        }
+    }
+}
+
+impl From<&'static str> for ImportEntity {
+    fn from(value: &'static str) -> Self {
+        match value {
+            "event" => Self::Event,
+            "event_price" => Self::EventPrice,
+            "gallery" => Self::Gallery,
+            "hall" => Self::Hall,
+            "location" => Self::Location,
+            "media" => Self::Media,
+            "media_variant" => Self::MediaVariant,
+            "price" => Self::Price,
+            "price_rank" => Self::PriceRank,
+            "production" => Self::Production,
+            "space" => Self::Space,
+            other => panic!("unknown import entity: {other}"),
+        }
+    }
+}
+
+/// Specifies what field causes problems
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ImportField {
+    Amount,
+    CdnUrl,
+    Crop,
+    Event,
+    Format,
+    GalleryUrl,
+    Location,
+    MaxTicketsPerOrder,
+    Name,
+    OpenSeating,
+    Price,
+    Production,
+    Rank,
+    SeatSelection,
+    SourceId,
+    SourceUri,
+}
+
+impl ImportField {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Amount => "amount",
+            Self::CdnUrl => "cdn_url",
+            Self::Crop => "crop",
+            Self::Event => "event",
+            Self::Format => "format",
+            Self::GalleryUrl => "gallery_url",
+            Self::Location => "location",
+            Self::MaxTicketsPerOrder => "max_tickets_per_order",
+            Self::Name => "name",
+            Self::OpenSeating => "open_seating",
+            Self::Price => "price",
+            Self::Production => "production",
+            Self::Rank => "rank",
+            Self::SeatSelection => "seat_selection",
+            Self::SourceId => "source_id",
+            Self::SourceUri => "source_uri",
+        }
+    }
+}
+
+impl From<&'static str> for ImportField {
+    fn from(value: &'static str) -> Self {
+        match value {
+            "amount" => Self::Amount,
+            "cdn_url" => Self::CdnUrl,
+            "crop" => Self::Crop,
+            "event" => Self::Event,
+            "format" => Self::Format,
+            "gallery_url" => Self::GalleryUrl,
+            "location" => Self::Location,
+            "max_tickets_per_order" => Self::MaxTicketsPerOrder,
+            "name" => Self::Name,
+            "open_seating" => Self::OpenSeating,
+            "price" => Self::Price,
+            "production" => Self::Production,
+            "rank" => Self::Rank,
+            "seat_selection" => Self::SeatSelection,
+            "source_id" => Self::SourceId,
+            "source_uri" => Self::SourceUri,
+            other => panic!("unknown import field: {other}"),
+        }
+    }
+}
+
+/// Specifies which linked entity is involved
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ImportRelation {
+    Event,
+    EventPrice,
+    Gallery,
+    Hall,
+    Location,
+    Media,
+    MediaVariant,
+    Price,
+    PriceRank,
+    Production,
+    Space,
+}
+
+impl ImportRelation {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Event => "event",
+            Self::EventPrice => "event_price",
+            Self::Gallery => "gallery",
+            Self::Hall => "hall",
+            Self::Location => "location",
+            Self::Media => "media",
+            Self::MediaVariant => "media_variant",
+            Self::Price => "price",
+            Self::PriceRank => "price_rank",
+            Self::Production => "production",
+            Self::Space => "space",
+        }
+    }
+}
+
+impl From<&'static str> for ImportRelation {
+    fn from(value: &'static str) -> Self {
+        match value {
+            "event" => Self::Event,
+            "event_price" => Self::EventPrice,
+            "gallery" => Self::Gallery,
+            "hall" => Self::Hall,
+            "location" => Self::Location,
+            "media" => Self::Media,
+            "media_variant" => Self::MediaVariant,
+            "price" => Self::Price,
+            "price_rank" => Self::PriceRank,
+            "production" => Self::Production,
+            "space" => Self::Space,
+            other => panic!("unknown import relation: {other}"),
+        }
+    }
+}
+
+/// Enum for importer-level failures, like S3 or IO issues, or network issues.
+/// These are run-ending failures.
 #[derive(Debug, Error)]
 pub enum ImporterError {
     #[error("Reqwest error: {0}")]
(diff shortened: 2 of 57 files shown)


Kaya-Sem:
fixes                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                            
- regenerated OpenAPI types (you); doc-commented productions.upsert_import field policy; added unresolved/resolved tab toggle in CMS table (types, hook, query key, i18n).                                                                                
- added 19 unit tests covering to_create error paths for space/hall/event + full to_import_error mapping for ImportItemError/ImportItemWarning. Replaced per-item resolve_for_item with batched resolve_for_items (one UPDATE per page instead of per item) across all five entity loops.                                                                                                                                                                                                                                                                                     
- fixed error.rs typo + dangling doc comment; hall slug now refuses to fabricate <name>-0 and returns MissingRequiredField("source_id"); removed hardcoded "skipping" from record_item_error; restored import-order comments in update_since_last; aligned LocationRepo::by_source_id to return Option like sibling repos.

Kaya-Sem (approved):

backend/api/src/models/event.rs
```diff
@@ -51,8 +52,24 @@ impl ApiEvent {
...
+        production_id: Uuid,
+        hall_id: Option<Uuid>,
+    ) -> Result<ItemConversion<EventCreate>, ImportItemError> {
+        let max_tickets_per_order = self
+            .max_tickets_per_order
+            .map(i32::try_from)
+            .transpose()
+            .map_err(|_| {
+                ImportItemError::out_of_range(
+                    "event",
```
hanneskt: suggestion: use an enum for this

backend/api/src/models/event.rs
```diff
@@ -51,8 +52,24 @@ impl ApiEvent {
...
+                    "max_tickets_per_order",
+                    self.max_tickets_per_order.unwrap_or_default().to_string(),
```
hanneskt:
```suggestion
                    format!("{:?}", self.max_tickets_per_order),
```

unwrap_or_default loses information: it gives a default value instead of the option context
`:?` uses the debug print implementation (`Debug` trait) which prints the option as well

backend/api/src/models/hall.rs
```diff
@@ -38,17 +38,23 @@ pub struct ApiHall {
...
         let seat_selection = match self.seat_selection.as_str() {
             "1" => true,
             "" => false,
             other => {
-                warn!(
-                    "unexpected seat_selection value: {}, defaulting to false",
-                    other
-                );
+                warnings.push(ImportItemWarning::UnexpectedFieldValue {
+                    entity: "hall",
```
hanneskt: suggestion: use an enum for this

backend/api/src/error.rs
```diff
@@ -11,3 +13,432 @@ pub enum ImporterError {
     #[error("S3 error: {0}")]
     S3(String),
 }
+
+/// Errors are modelled as explicit enums so we don't end up matching on hardcoded
```
hanneskt:
```suggestion
/// Errors are modeled as explicit enums so we don't end up matching on hardcoded
```

It's correct, but I think we use american english mostly? and this is uk english.
ultimately inconsequential

backend/api/src/error.rs
```diff
@@ -11,3 +13,432 @@ pub enum ImporterError {
...
+    pub warnings: Vec<ImportItemWarning>,
+}
+
+impl<T> ItemConversion<T> {
+    pub fn clean(value: T) -> Self {
+        Self {
+            value,
+            warnings: Vec::new(),
+        }
+    }
```
hanneskt: question: what is clean? why is there a clean?

Maybe name it `without_warnings` (or `empty`)? Maybe add a docstring if it has a special meaning.
I got a bit confused as to what `clean` meant when reading through the code

backend/api/src/models/space.rs
```diff
@@ -36,15 +37,64 @@ pub struct ApiSpace {
 // spaces are linked to a location, many_to_one. function that takes an ApiModel and a Uuid for
 // location, and returns a create model for Spaces
 impl ApiSpace {
-    pub fn to_create(self, location_uuid: Uuid) -> SpaceCreate {
+    pub fn to_create(
+        self,
+        location_uuid: Uuid,
+    ) -> Result<ItemConversion<SpaceCreate>, ImportItemError> {
```
hanneskt: suggestion: remove the wrap with `ItemConversion`

You're telling users of this function that it can return warnings, when in fact it can not.
Think of it like returning a `Result<T, E>` when you the function doesn't and can't produce errors. You're forcing handling of possible errors (here warnings) when there can not be any.

See the function signature as the documentation of what a function can return.

backend/api/src/models/space.rs
```diff
@@ -36,15 +37,64 @@ pub struct ApiSpace {
...
+        let space = make_space(ApiLocalizedText {
+            nl: None,
+            en: Some("Hall One".into()),
+            fr: None,
+        });
+
+        let conversion = space.to_create(Uuid::nil()).expect("conversion succeeds");
+        assert_eq!(conversion.value.name_nl, "Hall One");
+        assert!(conversion.warnings.is_empty());
+    }
```
hanneskt: issue: this test is weird

You're kind of testing functionality of an underlying function `flatten_single`, this test should go there (and be shorter and more focussed)

You're also not really testing what the test name says you are: you're testing if it falls back to english if there is no dutch translation, not if it prefers nl

fix: remove this test and test `flatten_single` properly

backend/api/src/error.rs
```diff
@@ -11,3 +13,432 @@ pub enum ImporterError {
...
+}
+
+#[derive(Debug, Error)]
+pub enum ImportItemError {
```
hanneskt, line 217: suggestion: some doc comments `///` on each of these would make using them a lot nicer

specifically *why* each variant should be used, and not another one

backend/api/src/lib.rs
```diff
@@ -132,19 +138,35 @@ impl ApiImporter {
...
+            let request = self.client.get(&url).query(&[
+                ("page", page.to_string().as_str()),
+                ("updated_at[after]", updated_after),
+            ]);
+
+            let response = match request
+                .header("X-AUTH-TOKEN", &self.auth_token)
+                .send()
+                .await
+            {
```
hanneskt:
```suggestion
            let request = self
                .client
                .get(&url)
                .query(&[
                    ("page", page.to_string().as_str()),
                    ("updated_at[after]", updated_after),
                ])
                .header("X-AUTH-TOKEN", &self.auth_token);

            let response = match request.send().await {
```

backend/api/src/lib.rs
```diff
@@ -166,7 +188,73 @@ impl ApiImporter {
         })
     }
 
-    pub async fn update_productions(&self, updated_after: &str) -> Result<(), reqwest::Error> {
+    async fn record_item_error(
```
hanneskt, line 205: suggestion: add some doc strings

backend/src/handlers/import_error.rs
```diff
@@ -0,0 +1,74 @@
...
+        (status = 200, description = "Success", body = PaginatedResponse<ImportErrorPayload>),
+        (status = 401, description = "Unauthorized", body = ErrorResponse)
+    ),
+    security(
+        ("cookie_auth" = [])
+    )
+)]
+pub async fn get_all(
+    db: Database,
+    Query(params): Query<GetImportErrorsQuery>,
```
hanneskt: suggestion: use the [paging query](https://github.com/SELab-2/viernulvier-6/blob/main/backend/src/handlers/queries/pagination.rs) on the main branch

backend/src/handlers/import_error.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let next_cursor = if errors.len() as i64 > limit {
+        errors.pop();
+        Some(encode_cursor(offset + limit))
+    } else {
+        None
+    };
+
+    Ok(Json(PaginatedResponse {
+        data: errors.into_iter().map(ImportErrorPayload::from).collect(),
```
hanneskt: suggestion: do this in function implemented on `ImportErrorPayload`

just like all the other handlers

backend/src/handlers/import_error.rs
```diff
@@ -0,0 +1,74 @@
...
+        next_cursor: next_cursor.map(Into::into),
+    }))
+}
+
+fn decode_cursor(cursor: Option<&str>) -> Option<i64> {
+    let cursor = cursor?;
+    let decoded = STANDARD_NO_PAD.decode(cursor).ok()?;
+    let offset = String::from_utf8(decoded).ok()?;
+    offset.parse::<i64>().ok()
+}
```
hanneskt: suggestion: look at how this is done in the productions paging ;)

hanneskt (changes requested):
General comments
- some more doc strings on new functions and enums would be nice
- maybe it's time to split the lib.rs? have a separate file for each imported object (you can have multiple `impl` blocks in multiple files)
- use more enums instead of strings when the variants are static and known beforehand
- pretty nice implementation overall!
- run `cargo fmt` :D

hanneskt:
You'll have conflict with this and #289 , probably worth it to have a look already

backend/api/src/error.rs
brentvent: Changed this to ```without_warnings```, it's basically just a conversion that did not produce any warnings.

backend/api/src/models/space.rs
brentvent: Tests are mostly AI generated. Sorry for having you read through slop! Bad oversight on my part.

backend/api/src/error.rs
brentvent: good idea, it's not very readable atm

frontend/src/types/api/import-error.api.types.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export interface ImportErrorsListResponse {
+    data: ImportErrorResponse[];
+    next_cursor: string | null;
+}
+
+export interface ImportErrorsListOptions {
+    pagination?: PaginationParams;
+    resolved?: boolean;
+}
```
Thoosje: These types should be received from the `generated.ts` types. This makes sure the frontend breaks if the logic of the backend changes. E.g. `PaginatedResponse_ImportErrorPayload` from `generated.ts`

Take a look at the other /types/api/...

frontend/src/hooks/api/query-keys.ts
```diff
@@ -12,6 +12,12 @@ const buildQueryKey = (
...
     version: ["version"] as const,
+    importErrors: {
+        all: (pagination?: PaginationParams, resolved?: boolean) => {
+            const base: readonly unknown[] = ["import-errors", { resolved: resolved ?? false }];
+            return pagination ? [...base, pagination] : base;
+        },
+    },
     locations: {
         all: (pagination?: PaginationParams) => buildQueryKey(["locations"], pagination),
         detail: (id: string) => ["locations", id] as const,
```
Thoosje, line 21: Can't you use something like in `locations` for `importErrors` and use `buildQueryKey(["import-errors"], pagination)`? Probably need to extend `buildQueryKey` to also accept `resolved`.

Thoosje (changes requested):
Only checked frontend: make sure to look at the comments as it's important to have the backend and frontend "connected"

backend/api/src/error.rs
```diff
@@ -1,5 +1,192 @@
...
+    Gallery,
+    Hall,
+    Location,
+    Media,
+    MediaVariant,
+    Price,
+    PriceRank,
+    Production,
+    Space,
+    Other(&'static str),
```
hanneskt: question: why other? if we are adding a new entity to import should we not add it to the enum also?

When would we ever use other?

backend/api/src/error.rs
```diff
@@ -1,5 +1,192 @@
...
+    MaxTicketsPerOrder,
+    Name,
+    OpenSeating,
+    Price,
+    Production,
+    Rank,
+    SeatSelection,
+    SourceId,
+    SourceUri,
+    Other(&'static str),
```
hanneskt: question: same here i suppose, although I can see more why you would use it here, but still the same question: why not force addding a new field to the enum?

backend/api/src/error.rs
```diff
@@ -11,3 +198,598 @@ pub enum ImporterError {
...
+        }
+    }
+
+    fn field(&self) -> Option<&'static str> {
+        match self {
+            Self::InvalidReference { field, .. }
+            | Self::MissingRequiredField { field, .. }
+            | Self::OutOfRange { field, .. } => Some(field.as_str()),
+            Self::ImportFailure { field, .. } => field.map(ImportField::as_str),
+            _ => None,
```
hanneskt: suggestion: handle all cases explicitly

if someone ever adds a case, the compiler will tell them this match doesn't cover all cases. With a `_` catchall, you rely on the person remembering/knowing they need to change this

backend/api/Cargo.toml
```diff
@@ -24,7 +24,7 @@ slug = "0.1.6"
 sha2 = "0.10"
 hex = "0.4"
 aws-sdk-s3 = { version = "1", features = ["behavior-version-latest"] }
-tokio = { version = "1.50.0", features = ["fs", "io-util", "rt"] }
+tokio = { version = "1.50.0", features = ["fs", "io-util", "rt", "time"] }
```
hanneskt: issue: we use `chrono` everywhere else, so it seems weird to use the `time` crate here?

backend/database/src/repos/import_error.rs
```diff
@@ -0,0 +1,187 @@
...
+                "SELECT *
+                 FROM import_errors
+                 WHERE {base_where}
+                 ORDER BY last_seen_at DESC, id DESC
+                 LIMIT $1"
+            ))
+            .bind(limit as i64)
+            .fetch_all(self.db)
+            .await?),
+        }
```
hanneskt: suggestion(non-blocking): you could use a querybuilder here to make it shorter/cleaner

but it's short enough that it doesn't matter that much

hanneskt (commented):
Image 😭

backend/api/src/error.rs
brentvent: yes, essentially a fallback, but if we're adding a new entity we're adding it to the enum no matter what, so I might as well remove it.

backend/api/Cargo.toml
brentvent: I use tokio time's ```sleep``` function for the exponential backoff inside ```lib.rs```

backend/database/src/repos/location.rs
```diff
@@ -204,25 +204,37 @@ impl<'a> LocationRepo<'a> {
...
+                existing.street = location.street;
+                existing.number = location.number;
+                existing.postal_code = location.postal_code;
+                existing.city = location.city;
+                existing.country = location.country;
+                existing.phone_1 = location.phone_1;
+                existing.phone_2 = location.phone_2;
+                existing.is_owned_by_viernulvier = location.is_owned_by_viernulvier;
+                existing.uitdatabank_id = location.uitdatabank_id;
+                existing.slug = location.slug;
```
hanneskt: question: could you do the same by setting the id of `location` here to the one from `existing`?

Could save you a lot of lines :D

backend/database/src/repos/location.rs
```diff
@@ -204,25 +204,37 @@ impl<'a> LocationRepo<'a> {
...
-        source_id: i32,
-    ) -> Result<Option<LocationWithTranslations>, DatabaseError> {
-        let Some(location) = Location::select()
-            .where_("source_id = $1")
-            .bind(source_id)
-            .fetch_optional(self.db)
-            .await?
-        else {
-            return Ok(None);
+    pub async fn upsert_import(&self, location: LocationCreate) -> Result<Location, DatabaseError> {
```
hanneskt: question: when is there ever not a `source_id`?

I assume this is only used for import (since the function is called `upsert_import`) so with import this seems like it couldn't happen.

Also, why is this function called `upsert_import`? (the import part)
edit: ah, it's because it's by `source_id`. I think maybe calling it `upsert_by_source_id` is a lot clearer

backend/database/src/repos/hall.rs
```diff
@@ -123,6 +123,27 @@ impl<'a> HallRepo<'a> {
...
+            Some(mut existing) => {
+                existing.slug = hall.slug;
+                existing.vendor_id = hall.vendor_id;
+                existing.box_office_id = hall.box_office_id;
+                existing.seat_selection = hall.seat_selection;
+                existing.open_seating = hall.open_seating;
+                existing.name = hall.name;
+                existing.remark = hall.remark;
+                existing.space_id = hall.space_id;
+                self.update(existing).await
```
hanneskt: same questions as for locations repo

backend/database/src/repos/event.rs
```diff
@@ -78,6 +74,37 @@ impl<'a> EventRepo<'a> {
...
+                existing.vendor_id = event.vendor_id;
+                existing.box_office_id = event.box_office_id;
+                existing.uitdatabank_id = event.uitdatabank_id;
+                existing.max_tickets_per_order = event.max_tickets_per_order;
+                existing.production_id = event.production_id;
+                existing.status = event.status;
+                existing.hall_id = event.hall_id;
+                self.update(existing).await
+            }
+            None => self.insert(event).await,
```
hanneskt: same questions as for locations repo

backend/database/src/repos/production/mod.rs
```diff
@@ -61,6 +61,38 @@ impl<'a> ProductionRepo<'a> {
...
+            Some(mut existing) => {
+                existing.production.slug = production.slug;
+                existing.production.video_1 = production.video_1;
+                existing.production.video_2 = production.video_2;
+                existing.production.eticket_info = production.eticket_info;
+                existing.production.uitdatabank_theme = production.uitdatabank_theme;
+                existing.production.uitdatabank_type = production.uitdatabank_type;
+                self.update(existing.production, translations).await
+            }
+            None => self.insert(production, translations).await,
```
hanneskt: same as for locations repo

backend/api/src/error.rs
hanneskt: You resolved and said you were removing it but didn't :D

backend/src/handlers/import_error.rs
```diff
@@ -0,0 +1,46 @@
...
+    error::ErrorResponse,
+    handlers::{IntoApiResponse, JsonResponse, queries::pagination::PaginationQuery},
+};
+use database::Database;
+
+#[derive(Deserialize, IntoParams)]
+pub struct GetImportErrorsFilterQuery {
+    #[serde(default)]
+    pub resolved: bool,
+}
```
hanneskt: put this in `handlers/queries`

backend/src/handlers/import_error.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[utoipa::path(
+    method(get),
+    path = "/import-errors",
+    operation_id = "get_import_errors",
+    tag = "Editor",
+    description = "List importer errors for CMS usage. Returns unresolved errors by default.",
+    params(PaginationQuery, GetImportErrorsFilterQuery),
```
hanneskt, line 19: question: is this needed?

hanneskt (changes requested):
Really good work btw!

You're almost there, just some small things. Feel free to push back if you have another opinion

backend/src/handlers/import_error.rs
brentvent: I tried to follow the other endpoints as much as possible. What do you mean exactly?

backend/api/src/error.rs
```diff
@@ -1,5 +1,186 @@
...
+            Self::OpenSeating => "open_seating",
+            Self::Price => "price",
+            Self::Production => "production",
+            Self::Rank => "rank",
+            Self::SeatSelection => "seat_selection",
+            Self::SourceId => "source_id",
+            Self::SourceUri => "source_uri",
+        }
+    }
+}
```
Thoosje, line 103: Why specific only these fields?

Thoosje (approved):
LGTM, 1 thing to think about: should the editor also have access to these errors or only admin?

Kaya-Sem:
> LGTM, 1 thing to think about: should the editor also have access to these errors or only admin?

I'd say editors too. The more eyes on the errors, the better

brentvent:
> > LGTM, 1 thing to think about: should the editor also have access to these errors or only admin?
> 
> I'd say editors too. The more eyes on the errors, the better

Agreed

backend/api/src/error.rs
brentvent: The enum contains the fields for which the importer currently reports explicit field-level validation errors or warnings. It can be expanded when we introduce more field-specific import validation.

Kaya-Sem:
3 weeks old already :sob:

backend/src/handlers/import_error.rs
hanneskt: The `params(PaginationQuery, GetImportErrorsFilterQuery),` line (github sometimes just adds more lines to review comments)

I think utoipa adds these automatically (I think other endpoints work like that?)

hanneskt: This one

hanneskt (approved):

PR #272, opened 2026-04-10
feat(production-page): Connected media hooks to production detail page

Thoosje:
**Description**
This PR connects the media hooks to the production page to display images. Added missing translations where needed.

**How Has This Been Tested?**
- [x] Local manual testing


backend/src/dto/production.rs
```diff
@@ -171,7 +171,7 @@ pub struct ProductionPayload {
     pub translations: Vec<ProductionTranslationPayload>,
 
     /// Cover image URL resolved from the entity_media link (output-only).
-    #[serde(skip_deserializing)]
+    #[serde(default)]
```
hanneskt, line 174: issue: by changing this it now shows up in the swagger payload

Image

it's not used on PUT so this is wrong. Why was it changed?

hanneskt (changes requested):

backend/src/dto/production.rs
Thoosje: With the other value it was no where visible.. other fixes did not work, I'm not sure how else to fix this.

Kaya-Sem: I believe in you

hanneskt: maybe split it into 2 structs?

Thoosje: Sure, I did not notice it was also in the put request

frontend/src/components/productionpage/production-related.tsx
```diff
@@ -1,56 +1,90 @@
+"use client";
+
+import Image from "next/image";
+import { useTranslations } from "next-intl";
+
 import { Link } from "@/i18n/routing";
 import { getLocalizedField } from "@/lib/locale";
+import { useGetEntityMedia } from "@/hooks/api/useMedia";
```
Kaya-Sem, line 8: doesn't production already contain the cover URL, thus making the useGetEntityMedia a bit redundant?

Thoosje: Yes/no, the list endpoint only provides the cover_url but the official 404 api supports different type of media (at the moment we only download the cover image but we can choose other images). The `useGetEntityMedia ` returns all the media associated with an entity.

Kaya-Sem (approved):

hanneskt (approved):

PR #283, opened 2026-04-12
fix: auth redirection

Kaya-Sem:
Fix CMS auth redirect so unauthenticated users land on `/login`.

- Rename `frontend/middleware.ts` to `frontend/src/proxy.ts`. Next.js 16 deprecated the `middleware` convention in favor of `proxy`, and with a `src/app` project the file has to live under `src/` or it isn't registered. Maybe this again doesn't work for tristan anymore, but thats a sacrifice we must make
- Drop the `/login → /cms` auto-redirect that fired when an `access_token` cookie was present. It caused a loop when the cookie was stale or invalid.
- Use `NextResponse.redirect(new URL(...))` so the configured `basePath` is preserved in preview deployments.
- In `api-client.ts`, when the refresh call fails on a `/cms` page, hard-navigate to `/login`. Covers the stale-but-present-cookie case that the proxy can't detect.
- Add `getBasePath()` helper to prefix client-side redirects with the preview path (axios interceptor runs outside React, so `useRouter` isn't available).
- Backend: `logout` was clearing `refresh_token` at cookie path `/refresh`, but login sets it at `/auth/refresh`. The mismatched path meant the cookie was never actually cleared.

**Related Issue**
Fixes #

**How Has This Been Tested?**
- [ ] Local manual testing

**Developer Checklist:**


Thoosje:
> Rename frontend/middleware.ts to frontend/src/proxy.ts. Next.js 16 deprecated the middleware convention in favor of proxy, and with a src/app project the file has to live under src/ or it isn't registered.

This is correct, it used to be proxy but remember that @tst-038 changed it to middleware as safari couldn't login otherwise.

Kaya-Sem:
> > Rename frontend/middleware.ts to frontend/src/proxy.ts. Next.js 16 deprecated the middleware convention in favor of proxy, and with a src/app project the file has to live under src/ or it isn't registered.
> 
> This is correct, it used to be proxy but remember that @tst-038 changed it to middleware as safari couldn't login otherwise.

see my update. He should try see if its fixed now

Thoosje (approved):

hanneskt (approved):

tst-038:
Doesn't work on safari...

Kaya-Sem:
> Doesn't work on safari...

de middleware stond wel gwn fout en werd niet geregistreerd. Mss middleware gebruiken, maar wel op juiste plek, is een gulden tussenweg?

tst-038:
Summary of changes:
  1. Frontend base-path — getBasePath() now uses NEXT_PUBLIC_BASE_PATH instead of fragi
     window.location parsing.
  2. Backend cookie config — added two new env vars: COOKIE_SECURE (default true) and COOKIE_SAME_SITE (default strict).
  3. New session_present cookie — a safe boolean flag on path=/ that lets the Next.js proxy know a session exists, without exposing the actual refresh-token secret.
  4. refresh_token path stays restricted — back to /api/auth/refresh (or /{preview}/api
     th/refresh) as intended for security.
  5. Middleware fix (proxy.ts) — only redirects to /login when both access_token and se
     on_present are missing. This gives the axios interceptor a chance to refresh an ex
     ed token.
  6. Docs & env examples updated.

  Why: in dev, Safari chokes on cross-origin cookies (localhost:3000 ↔ 3001). These changes make login and token refresh work reliably in Safari without window.location hack

Kaya-Sem:
> Summary of changes:
> 
>     1. Frontend base-path — getBasePath() now uses NEXT_PUBLIC_BASE_PATH instead of fragi
>        window.location parsing.
> 
>     2. Backend cookie config — added two new env vars: COOKIE_SECURE (default true) and COOKIE_SAME_SITE (default strict).
> 
>     3. New session_present cookie — a safe boolean flag on path=/ that lets the Next.js proxy know a session exists, without exposing the actual refresh-token secret.
> 
>     4. refresh_token path stays restricted — back to /api/auth/refresh (or /{preview}/api
>        th/refresh) as intended for security.
> 
>     5. Middleware fix (proxy.ts) — only redirects to /login when both access_token and se
>        on_present are missing. This gives the axios interceptor a chance to refresh an ex
>        ed token.
> 
>     6. Docs & env examples updated.
> 
> 
> Why: in dev, Safari chokes on cross-origin cookies (localhost:3000 ↔ 3001). These changes make login and token refresh work reliably in Safari without window.location hack

works for me on firefox :+1:

Issue #286, opened 2026-04-12
feat(backend): stage 2 - seed binary runs normalization on cached JSON and produces pg_dump


Kaya-Sem:
## Parent

Part of #253 - AI-powered data normalization pipeline.
Stage 2 of the offline seed pipeline.

Depends on:
- #254 (LLM client infrastructure)
- #255 (artist extraction)
- #256 (genre classification)
- #257 (location name enrichment)
- Importer refactor issue (split HTTP fetch from DB insert)
- Stage 1 fetch binary (raw JSON mirror)

## Summary

New offline binary that reads the cached raw 404 API JSON (from stage 1), runs deterministic pre-pass and LLM normalization, and writes the fully normalized result to a fresh Postgres. Output: `backend/seed/seed.sql`, committed via Git LFS and restored on every environment's first boot.

### Binary

- New bin target: `backend/api/src/bin/seed.rs`
- Requires `ai-normalization` cargo feature
- Invocation: `cargo run --release -p api --bin seed --features ai-normalization`
- Uses `SEED_DATABASE_URL` env var (distinct from dev `DATABASE_URL`) to avoid clobbering local work
- Reads `LLM_PROVIDER`, `LLM_API_KEY`, `LLM_MODEL`, `LLM_CONFIDENCE_THRESHOLD` from env

### Flow

1. Connect to the seed DB, run all migrations
2. Read `backend/seed/raw/*.json` into the parsed API structs (shared with the importer after the split-fetch-from-insert refactor)
3. Feed them into the shared insert path
4. For every production, call `normalize_production` (shared module from #254). Apply actions via existing repos. Write `normalization_log` rows. Set `productions.normalized_at = now()` on success
5. Set `internal_state.LastApiUpdate` to `manifest.json.max_updated_at` so downstream envs resume deltas from exactly the cut point
6. Exit cleanly

### Dump workflow

- Operator runs, outside the binary:
  `pg_dump --data-only --no-owner --no-privileges --disable-triggers "$SEED_DATABASE_URL" > backend/seed/seed.sql`
- `backend/seed/seed.sql` is tracked by Git LFS (`.gitattributes`: `backend/seed/*.sql filter=lfs diff=lfs merge=lfs -text`)
- Document the full regenerate-seed flow in `backend/README.md` (stage 1, stage 2, pg_dump, commit)

### Rerun scenarios

- Bump normalizer version: rerun stage 2 only against the committed raw JSON. No 404 API calls.
[...]

PR #289, opened 2026-04-12
chore(backend): split importer and fetch

### LINKED ISSUE #284: chore(backend): split ApiImporter HTTP fetch from DB insert
## Parent

Part of #253 - AI-powered data normalization pipeline.
Prerequisite for the stage 2 seed binary.

## Summary

Refactor `backend/api/src/lib.rs` so the insert-entities-into-DB logic is separated from the HTTP fetch. Both the live importer and the new offline seed binary need to share the insert path.

## Motivation

The seed binary (stage 2 of the normalization pipeline) reads cached 404 API responses from local JSON files rather than calling the API, then inserts them into Postgres using the same logic the live importer uses. Today that logic is interleaved with the HTTP client calls inside `ApiImporter`.

## Requirements

- Extract the per-entity insert flow (productions, genres, locations, spaces, halls, events, prices, price_ranks, event_prices, media galleries) into a module that takes an iterator of already-parsed API structs plus a DB pool.
- The existing `ApiImporter::update_since_last` continues to work unchanged from the outside: it fetches over HTTP, then feeds the parsed structs into the extracted insert path.
- A future seed binary can construct the same parsed structs from on-disk JSON and feed them in directly.
- No behavior change. Existing tests pass.

## Acceptance criteria

- [ ] Insert logic is callable without an HTTP client
- [ ] `ApiImporter::update_since_last` still works end to end
- [ ] Raw API response types are `serde::Deserialize` so they can be loaded from JSON files
- [ ] No change to what gets written to the DB for a given input

Kaya-Sem:
Extract per-entity DB insert logic from `ApiImporter` into a new `api::insert` module so both the live HTTP importer and the upcoming offline seed binary (#286) can share it.

- New `backend/api/src/insert/{locations,spaces,halls,productions,events,prices,price_ranks,event_prices}.rs`, one async fn per entity taking `&Database` + a parsed `Api*` struct.
- `update_*` methods in `ApiImporter` shrink to stream-loop + per-item delegate call; S3 gallery logic stays on the live importer (seed binary does not touch S3).
- `insert_production` returns the source_id so the live path can still drive gallery fetches; seed path ignores it.
- Behavior preserved: same lookups, same warn-and-skip on missing FKs, same upsert semantics for prices / price_ranks / event_prices.

**Related Issue**
Fixes #284

**How Has This Been Tested?**
- [ ] Local manual testing

**Developer Checklist:**

### DIFF
--- a/backend/api/src/insert/events.rs
+++ b/backend/api/src/insert/events.rs
@@ -1,0 +1,30 @@
+use database::{Database, error::DatabaseError};
+use tracing::warn;
+
+use crate::models::event::ApiEvent;
+
+pub async fn insert_event(db: &Database, event: ApiEvent) -> Result<(), DatabaseError> {
+    let Some(prod_source_id) = event.production_source_id() else {
+        warn!("Events: event has no production source_id, skipping");
+        return Ok(());
+    };
+
+    let Some(production) = db.productions().by_source_id(prod_source_id).await? else {
+        warn!("Events: production source_id {prod_source_id} not found in db, skipping");
+        return Ok(());
+    };
+
+    let hall_uuid = if let Some(hall_source_id) = event.hall_source_id() {
+        let hall_opt = db.halls().by_source_id(hall_source_id).await?;
+        if hall_opt.is_none() {
+            warn!("Events: hall source_id {hall_source_id} not found in db");
+        }
+        hall_opt.map(|h| h.id)
+    } else {
+        None
+    };
+
+    let event_create = event.to_create(production.production.id, hall_uuid);
+    db.events().insert(event_create).await?;
+    Ok(())
+}
--- a/backend/api/src/insert/events.rs
+++ b/backend/api/src/insert/events.rs
@@ -3,28 +3,30 @@ use tracing::warn;
 
 use crate::models::event::ApiEvent;
 
-pub async fn insert_event(db: &Database, event: ApiEvent) -> Result<(), DatabaseError> {
-    let Some(prod_source_id) = event.production_source_id() else {
-        warn!("Events: event has no production source_id, skipping");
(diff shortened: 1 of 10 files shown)


backend/api/src/insert/events.rs
```diff
@@ -0,0 +1,30 @@
+use database::{Database, error::DatabaseError};
+use tracing::warn;
+
+use crate::models::event::ApiEvent;
+
+pub async fn insert_event(db: &Database, event: ApiEvent) -> Result<(), DatabaseError> {
```
hanneskt: suggestion: Implement these functions on the structs they are taking as a second argument

so here:
```rust
impl ApiEvent {
    pub async fn insert(db: &Database) ...
}
```

maybe even make it a trait for later? but I'm not certain if that would be useful in the future

hanneskt (changes requested):

hanneskt (approved):

Thoosje (approved):

backend/api/src/insert/productions.rs
```diff
@@ -0,0 +1,22 @@
...
+
+impl ApiProduction {
+    /// returns the 404 API source_id so the live importer can drive per-production
+    /// gallery fetches. seed binary ignores this.
+    pub async fn insert(self, db: &Database) -> Result<Option<i32>, DatabaseError> {
+        let source_id = self
+            .id
+            .split('/')
+            .next_back()
+            .and_then(|s| s.parse::<i32>().ok());
```
brentvent: suggestion: use the helper ```extract_source_id```

brentvent (changes requested):

brentvent (approved):

PR #291, opened 2026-04-12
feat: properly import event statuses

Kaya-Sem:
Resolve event status IRIs to display strings during import, instead of storing the raw IRI in the `status` column.

- Add `ApiEventStatus` model with a `display()` helper that prefers `short_name`, falls back to Dutch `name`, and tolerates either field being missing or empty.
- In `update_events`, drain `/events/statuses` once into an `@id -> display` map before iterating events, and pass the resolved string into `ApiEvent::to_create`. Events whose status IRI isn't in the map are skipped with a warning.
- Lower the default "last updated" fallback to `1900-01-01` so a first-time import pulls everything.

**Related Issue**
Fixes #

**How Has This Been Tested?**
- [ ] Local manual testing

**Developer Checklist:**


backend/api/src/lib.rs
```diff
@@ -279,16 +282,35 @@ impl ApiImporter {
...
         while let Some(batch_result) = stream.next().await {
             let events = batch_result?;
             let amt = events.len();
             info!("Events: got {amt} from api");
+            let mut inserted: usize = 0;
             for event in events {
-                event.insert(&self.db).await.unwrap();
+                if event.insert(&self.db, &status_map).await.unwrap() {
+                    inserted += 1;
+                }
```
tst-038: Not sure but won't this stop all remaining events from importing if 1 fails during the process?

hanneskt: yes. We should wait until #211 lands and use it's system to properly handle cases like this

hanneskt (commented):
Wait on #211 and use it's framework

hanneskt (approved):

Thoosje (approved):

PR #295, opened 2026-04-13
feat: preview for production, article edits

tst-038:
**Description**
 This PR introduces an iframe-based live preview system for the CMS, letting editors see draft changes for productions and articles in real time without publishing.
  Key changes:
  • Adds PreviewContext with localStorage persistence and cross-tab sync for preview state.
  • Introduces useArticlePreview, useProductionPreview, usePreviewData, and useMounted hooks.
  • Builds a dedicated production editor page with a live iframe preview pane, plus a production-converter utility to handle format conversion between table rows and API shapes.
  • Updates the article editor to use the same iframe preview flow via ?preview=1.
  • Adds a PreviewBadge component as a visual indicator when preview mode is active.
  • Extracts a shared LocaleSwitcherLinks component so switching languages (NL/EN) preserves the ?preview=1 query param and doesn't exit preview mode.
  • Adds a new Textarea UI component.
  • Updates related tests (article mapper, production mapper, ArchiveSidebar, and MSW handlers).


brentvent (approved):
Nice feature. Having 2 tabs open and simultaneously previewing and editing productions is a bit unreliable: whenever you change something in one preview, you clear the other preview. It's only a visual bug, the fields themselves do not change.

tst-038:
> Nice feature. Having 2 tabs open and simultaneously previewing and editing productions is a bit unreliable: whenever you change something in one preview, you clear the other preview. It's only a visual bug, the fields themselves do not change.

True thats not supported, I could modify the param to take an id which would support this

Kaya-Sem (changes requested):
as discussed on discord, the mobile situation should be look at. Also enabling preview by default might be desirable

Thoosje (commented):
Like @Kaya-Sem said: fix mobile, looks good.

Kaya-Sem:
suggestion(non-blocking): swap places for general and content. The content would be altered more i think and is better to find quickly. Now you have to look and find to find title and description etc

image

Kaya-Sem:
Dont know if to request changes or block for this, but the edit preview page for articles is basically unusable on mobile

tst-038:
> suggestion(non-blocking): swap places for general and content. The content would be altered more i think and is better to find quickly. Now you have to look and find to find title and description etc
> 
> image

ye maybe idk, just feels weird to put general underneath. @Thoosje @jarnex47 what do y'all think?

Kaya-Sem:
> suggestion(non-blocking): swap places for general and content. The content would be altered more i think and is better to find quickly. Now you have to look and find to find title and description etc
> image

I think this would be solved easily by renaming it to something else than "general", as the content is more general imo. "Other" or "miscellaneous" seems more fiiting.

Kaya-Sem (approved):

PR #303, opened 2026-04-14
fix: filters now sticky, searchbar appears when hero is scrolled away

Kaya-Sem:
Make the search page filters sticky and reveal the search input in the results bar once the hero scrolls out of view.

- Move `ResultsBar` into the main results column and make `ArchiveSidebar` sticky so filters stay in view while scrolling
- Add an `IntersectionObserver` on `SearchHero` to toggle a compact search input inside `ResultsBar` when the hero leaves the viewport
- Wire `SearchHero`'s query state through to the page (previously a no-op) and forward a ref via `forwardRef`
- Introduce a `(with-footer)` route group under `(app)/` so the footer only renders on content pages, not the search page, where it conflicts with infinite scroll
- Update `ResultsBar` tests for the new props

**Related Issue**
Fixes #

**How Has This Been Tested?**
- [ ] Local manual testing

**Developer Checklist:**


tst-038 (approved):

tst-038:
Love the changes but we should maybe discuss if we want our header to be sticky to, ...
Because of the infinite scroll this might be desirable as you wouldn't be required to scroll a bazillion productions up if you wanted to change the locale or smth (assuming people dont know they can change the URL ofc)

Kaya-Sem:
> Love the changes but we should maybe discuss if we want our header to be sticky to, ... Because of the infinite scroll this might be desirable as you wouldn't be required to scroll a bazillion productions up if you wanted to change the locale or smth (assuming people dont know they can change the URL ofc)

theres this trick we can do to circumvent this: many pages scroll away the headers, and you can indeed keep scrolling infinitely, but the moment you scroll back up even a little bit, the header reappears.

In addition to this, many sites also provide a button to navigate back to top

hanneskt (approved):

PR #312, opened 2026-04-16
fix(backend,frontend): validate uploads and prevent s3_key injection

tst-038:
# Fix S3 Upload Security Vulnerabilities

This PR addresses multiple security vulnerabilities related to S3 media uploads and metadata integrity.

---

### Problem
`POST /media/upload-url` accepted any user-supplied `filename` and `mime_type` without validation. An editor could request a presigned URL for `evil.html` with `text/html`, upload it to the publicly readable S3 bucket, and serve executable content — leading to XSS, malware distribution, or phishing.

### Fix
- **Extension whitelist**: only `jpg`, `jpeg`, `png`, `gif`, `webp`, `svg`, `mp4`, `pdf` are allowed.
- **MIME-type whitelist**: only matching image/video/PDF MIME types are accepted.
- **Extension ↔ MIME consistency check**: e.g. `.jpg` must be `image/jpeg`; mismatches are rejected.
- **Size limit**: enforced via `content_length` on the S3 presigned `put_object` request. The limit is configurable via `MAX_UPLOAD_SIZE_MIB` (default: `50`).

---

### Problem
`POST /media/entity/{type}/{id}/attach` inserted media with `ON CONFLICT (s3_key) DO UPDATE SET ...`. Because `s3_key` was user-controlled and exposed to editors, an attacker could:
1. Enumerate existing `s3_key` values from the public `GET /media` list.
2. Send an `attach` request reusing someone else's `s3_key`.
3. Overwrite that media's metadata (`alt_text`, `description`, `credit`, etc.) and forcibly link it to their own entity.

### Fix
- **Upload token (HMAC)**: `generate_upload_url` now returns an `upload_token` (keyed SHA-256 of the `s3_key`). `attach_to_entity` rejects any request without a valid token, proving the file was actually uploaded through the backend.
- **Existing-key rejection**: before insertion, `attach_to_entity` checks if the `s3_key` already exists in the database. If it does, the endpoint returns `409 Conflict`, making the `ON CONFLICT` hijack path impossible.

---

### `POST /api/media/entity/{type}/{id}/link`

Allows editors to link an **existing** media record to an entity (production, article, event, etc.) without uploading a new file. Intended for reusing old media, e.g. the same image for a new newspaper article.

- **Requires**: `media_id`, optional `role` / `sort_order` / `is_cover_image`
- Does **not** modify any metadata of the original media item
- Does **not** require an `upload_token`

This cleanly separates the **new upload** flow (`attach_to_entity`, with token + `s3_key` validation) from the **reuse existing media** flow (`link_to_entity`).

### DIFF
--- a/backend/src/handlers/media.rs
+++ b/backend/src/handlers/media.rs
@@ -5,7 +5,8 @@ use axum::{
     http::StatusCode,
 };
 use database::{Database, models::entity_type::EntityType};
-use sha2::{Digest, Sha256};
+use hmac::{Hmac, Mac};
+use sha2::Sha256;
 use std::collections::HashSet;
 use std::time::Duration;
 use uuid::Uuid;
@@ -146,16 +147,23 @@ fn validate_upload_request(req: &UploadUrlRequest) -> Result<(), AppError> {
     Ok(())
 }
 
+type HmacSha256 = Hmac<Sha256>;
+
 pub fn generate_upload_token(secret: &str, s3_key: &str) -> String {
-    let mut hasher = Sha256::new();
-    hasher.update(secret.as_bytes());
-    hasher.update(b":");
-    hasher.update(s3_key.as_bytes());
-    hex::encode(hasher.finalize())
+    let mut mac = HmacSha256::new_from_slice(secret.as_bytes())
+        .expect("HMAC can take key of any size");
+    mac.update(s3_key.as_bytes());
+    hex::encode(mac.finalize().into_bytes())
 }
 
 fn verify_upload_token(secret: &str, s3_key: &str, token: &str) -> bool {
-    generate_upload_token(secret, s3_key) == token
+    let Ok(expected) = hex::decode(token) else {
+        return false;
+    };
+    let mut mac = HmacSha256::new_from_slice(secret.as_bytes())
+        .expect("HMAC can take key of any size");
+    mac.update(s3_key.as_bytes());
+    mac.verify_slice(&expected).is_ok()
 }
 
 #[utoipa::path(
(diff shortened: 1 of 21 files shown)


Thoosje (approved):

backend/src/handlers/media.rs
```diff
@@ -93,6 +95,77 @@ pub async fn get_one(
...
+    }
+
+    Ok(())
+}
+
+type HmacSha256 = Hmac<Sha256>;
+
+pub fn generate_upload_token(secret: &str, s3_key: &str) -> String {
+    let mut mac = HmacSha256::new_from_slice(secret.as_bytes())
+        .expect("HMAC can take key of any size");
```
hanneskt: suggestion: use a result or a valid fallback, as the thread will crash if you do expect and it is an error.

Expect has the same behavior as unwrap, just with a message.
A result (and a corresponding AppError variant) will allow the backend to send a proper response to the user, instead of just closing the connection

backend/src/handlers/media.rs
```diff
@@ -93,6 +95,77 @@ pub async fn get_one(
...
+    mac.update(s3_key.as_bytes());
+    hex::encode(mac.finalize().into_bytes())
+}
+
+fn verify_upload_token(secret: &str, s3_key: &str, token: &str) -> bool {
+    let Ok(expected) = hex::decode(token) else {
+        return false;
+    };
+    let mut mac = HmacSha256::new_from_slice(secret.as_bytes())
+        .expect("HMAC can take key of any size");
```
hanneskt: here also

hanneskt (approved):
just maybe don't use expect, otherwise okay

PR #315, opened 2026-04-16
fix(infra): preview on Kaya-slave

Thoosje:
**Description**
Creates a new traefik docker-compose specific for Kaya-Slave without SSL as Kaya-Sem his server runs behind a cloudflare tunnel (provides SSL). With the original SSL configuration it kept failing.


Kaya-Sem:
is this still needed

Thoosje:
> is this still needed


Unless you made the changes elsewhere yes: this splits the traefik. Or does it work alreedy in other branches without this fix? If so the splitting is not needed.

Kaya-Sem (approved):

tst-038 (approved):
🥇

Issue #321, opened 2026-04-16
frontend: improve styling notifications


Kaya-Sem:
We should try and style the notifcations to better fit our theme:
match shape, better fitting icon, maybe a different shade of green

<img width="883" height="457" alt="Image" src="https://github.com/user-attachments/assets/6914f15a-5f82-4fef-9e75-d2f525825eb7" />

Kaya-Sem:
Done: toast notifications were restyled to match the project theme in PR #330.

PR #330, opened 2026-04-16
chore(ui): restyle toast notifications to match theme

Kaya-Sem:
Swaps the `CircleCheckIcon`/`OctagonXIcon` icons for simpler `Check`/`X` that fit the chip-sized toast better. Adds per-type left-border accents to differentiate toast types visually without introducing new color tokens: success gets a `foreground` border, error gets `destructive`, warning and info get `muted-foreground`. All background and text tokens continue to use the existing `--popover`/`--border` variables.

Refs #321

image

### DIFF
--- a/frontend/src/components/ui/sonner.tsx
+++ b/frontend/src/components/ui/sonner.tsx
@@ -29,7 +29,7 @@ const Toaster = ({ ...props }: ToasterProps) => {
                     "--error-bg": "var(--popover)",
                     "--error-text": "var(--destructive)",
                     "--error-border": "var(--destructive)",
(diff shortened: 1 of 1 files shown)


tst-038 (approved):

tst-038:
on second thought maybe, we should consider not using rounded edges, most of our styling is with harsh right angle turns. I think we should make up our minds and choose 1, I'd go for the sharp corners...

image
Something like this (from @jarnex47 work in progress article page redesign)
I'd just consider thickening up the bottom and right border to create this shadowy effect...

hanneskt (approved):
good, and I think we should go with tst's suggestion of using right angles

PR #332, opened 2026-04-16
fix: keep footer at bottom on short pages by wrapping app layout chil…

tst-038:
prevents footer from sometimes showing up in the middle of the screen when no content is displayed


Kaya-Sem:
lol have never seen this happen

tst-038:
> lol have never seen this happen

Your screen probably ain't high enough, only happens if I go full or near full height on my monitor (and was temporarily except for article page)

Kaya-Sem (approved):

Thoosje (approved):

PR #335, opened 2026-04-16
feat: show stats counts on CMS overview cards

tst-038:
Small feature introducing stats in cms overview page using the stats endpoint implemented by @jarnex47 

https://github.com/user-attachments/assets/f3a61f55-5321-48e8-a56b-00a3afab1ea1

### DIFF
--- a/frontend/src/app/[locale]/(cms)/cms/(overview)/page.tsx
+++ b/frontend/src/app/[locale]/(cms)/cms/(overview)/page.tsx
@@ -3,15 +3,7 @@
 import { useTranslations } from "next-intl";
 import { useEffect, useRef } from "react";
 import { animate, stagger } from "animejs";
-import {
-    FileUp,
-    Database,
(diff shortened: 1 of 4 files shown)


Kaya-Sem:
I think its a good addition, but we should reconsider some of our icons
- a zip for collections is not really the right signal. Perhaps more like a star, stacked cards, a bookmark icon, ...
- the production has the clapperboard icon, which ties it into video production alot. might need to reconsider

Thoosje (approved):

tst-038:
> I think its a good addition, but we should reconsider some of our icons
> 
> * a zip for collections is not really the right signal. Perhaps more like a star, stacked cards, a bookmark icon, ...
> * the production has the clapperboard icon, which ties it into video production alot. might need to reconsider

image
@Kaya-Sem

Kaya-Sem (approved):

PR #336, opened 2026-04-16
feat(backend): add collection search

### LINKED ISSUE #237: feat(api): search endpoint for collections
Add `GET /collections/search?q=` endpoint with text search on collection title. Returns paginated results.

Needed by #213 and #167.

jerjoen:
**Description**
Adds search endpoint for collections
**Related Issue**
Fixes #237 

**How Has This Been Tested?**
- [x] Local manual testing

**Developer Checklist:**


backend/database/src/repos/collection.rs
```diff
@@ -35,14 +37,117 @@ impl<'a> CollectionRepo<'a> {
...
+                        if c.id < cursor.id {
+                            filtered.push(c);
+                        }
+                    }
+                    collections = filtered;
+                }
+
+                let next_cursor = if collections.len() == limit as usize {
+                    collections.pop();
+                    collections.last().map(|c| CursorData {
```
Kaya-Sem: I think there's a possible bug here. The cursor is applied in rust after fetching from the database, but the LIMIT is already applied in SQL before the cursor filter runs. This would mean pages 2+ would (silently) return fewer items than requested. Could also maybe not give next_cursor even when more pages exist.

I think ths could be fixed by puttin cursor condition into the SQL query before limit, like productionrepo does it:
```rust
let mut query = QueryBuilder::new("SELECT * FROM collections WHERE 1=1");                                               
      if let Some(cursor) = cursor {                                                                                          
          query.push(" AND id < ").push_bind(cursor.id);                                                                      
      }                     
      query.push(" ORDER BY id DESC LIMIT ").push_bind(limit);  
```

backend/tests/collection.rs
```diff
@@ -19,8 +22,114 @@ async fn get_all(db: PgPool) {
...
+    );
+    println!(
+        "page3: {:?}",
+        page3
+            .data
+            .iter()
+            .map(|c| c.slug.clone())
+            .collect::<Vec<_>>()
+    );
+    println!("page3 cursor: {:?}", page3.next_cursor);
```
Kaya-Sem: These print statements should probably be removed before merging

Kaya-Sem:
otherwise seems good to me! A nice addition :partying_face:

backend/migrations/20260415142824_collections_gin_index.up.sql
```diff
@@ -0,0 +1,8 @@
+-- Add up migration script here
+ALTER TABLE collection_translations
+ADD COLUMN full_search_text text
+GENERATED ALWAYS AS (immutable_concat_ws(' ', title)) STORED;
```
hanneskt: issue: remove this generated column

you only need this to give the search query all the text it needs to search in 1 big block, if you have multiple columns to search. For 1 column, you can just use that column.

Still make the index, just on the `title` column instead of the generated column

backend/database/src/repos/collection.rs
hanneskt: I concur. If I thought about it correctly, it would only ever give page 1 with a cursor, then a single item on page 2.

Small improvement on Kaya-Sem: since there is only one condition, you can just push the where together with it and skip the ugly `1=1` :D

```rust
let mut query = QueryBuilder::new("SELECT * FROM collections ");                                               
      if let Some(cursor) = cursor {                                                                                          
          query.push(" WHERE id < ").push_bind(cursor.id);                                                                      
      }                     
      query.push(" ORDER BY id DESC LIMIT ").push_bind(limit);  
```

In general, try to do things in SQL if you can :)

backend/database/src/repos/collection.rs
```diff
@@ -35,14 +37,117 @@ impl<'a> CollectionRepo<'a> {
...
+
+                if let Some(cursor) = cursor {
+                    let mut filtered = Vec::new();
+                    for c in collections {
+                        if c.id < cursor.id {
+                            filtered.push(c);
+                        }
+                    }
+                    collections = filtered;
+                }
```
hanneskt, line 122: You shouldn't do it like this here anyway, but if you ever need to do something like this use

`collections.into_iter().filter(|c| c.id < cursor.id).collect()`

It's shorter, cleaner and can be faster

hanneskt (changes requested):

backend/database/src/repos/collection.rs
```diff
@@ -1,20 +1,22 @@
...
-            CollectionWithTranslations,
+            Collection, CollectionCreate, CollectionSearch, CollectionTranslation,
+            CollectionTranslationData, CollectionWithScore, CollectionWithTranslations,
         },
         collection_item::{
             CollectionItem, CollectionItemBulkUpdate, CollectionItemCreate,
             CollectionItemTranslation, CollectionItemTranslationData,
             CollectionItemWithTranslations,
         },
+        cursor::CursorData,
```
hanneskt:
```suggestion
    filtering::cursor::CursorData,
```
Did you rebase and forget to compile? :p

backend/src/dto/collection.rs
```diff
@@ -7,14 +8,19 @@ use database::{
             CollectionContentType, CollectionItemBulkUpdate, CollectionItemCreate,
             CollectionItemTranslationData, CollectionItemWithTranslations,
         },
+        cursor::CursorData,
```
hanneskt:
```suggestion
    filtering::cursor::CursorData,
```
Same here

backend/tests/collection.rs
```diff
@@ -19,8 +22,114 @@ async fn get_all(db: PgPool) {
...
+
+    let data: PaginatedResponse<CollectionPayload> = response.into_struct().await;
+
+    assert!(data.data.is_empty());
+    assert!(data.next_cursor.is_none());
+}
+
+#[sqlx::test(fixtures("collections"))]
+#[test_log::test]
+async fn get_search_paginated(db: PgPool) {
```
hanneskt, line 60: issue: add a test for pagination without a search query as well

This would have caught your paging bug :p

hanneskt (changes requested):

backend/database/src/repos/collection.rs
Kaya-Sem: "In general, try to do things in SQL if you can :)"  
See https://github.com/SELab-2/viernulvier-6/blob/main/backend/migrations/20260405000000_entity_facets_fn.up.sql

Kaya-Sem:
any updates on this? can this be re-reviewed?

jerjoen:
> any updates on this? can this be re-reviewed?

Yes, it's ready for re-review now. I was waiting until after the Milestone 2 deadline.

Kaya-Sem (approved):

hanneskt (approved):

Kaya-Sem:
fixed your merge conflicts...

PR #345, opened 2026-04-18
feat(cms): wire up entity search

### LINKED ISSUE #167: feat(cms): add searchbar
**Describe the solution you'd like**
In the manage content tab, a nice searchbar should be implemented, placed above the datatable.
Entities should be fuzzily searchable on all their metadata

Kaya-Sem:
Adds text search to the productions, locations, and articles CMS tables. Typing in the search bar filters the list in real time (300 ms debounce, `?q=` query param, cursor reset on each new query).

Backend: extracts a shared `search_articles` helper so the existing public endpoint keeps filtering to published-only, while a new `/articles/cms/search` editor route searches across all statuses.

Frontend: new `SearchInput` component sits in the action bar row on the left; bulk-action controls appear on the right when rows are selected. The productions and locations infinite queries are updated to forward the search param.

**Related Issue**
Fixes #167
image


Thoosje (approved):
Looks good, take a look at some comments of copilot

backend/database/src/repos/article.rs
```diff
@@ -190,7 +209,10 @@ impl<'a> ArticleRepo<'a> {
                 .push(" <<-> a.title AS distance_score");
         }
 
-        builder.push(" FROM articles a WHERE a.status = 'published'");
+        builder.push(" FROM articles a WHERE 1=1");
```
tst-038, line 213: Kinda smelly

tst-038 (approved):

backend/database/src/repos/article.rs
Kaya-Sem: dismissive gesture

PR #358, opened 2026-04-20
fix(cms): add collapse-all button for expanded table rows

### LINKED ISSUE #325: feat(cms): collapse all button
there is currently no way to collapse all expanded selections, which is annoying. only refresh works

Image

tst-038:
- DataTable now accepts optional expanded/onExpandedChange props for controlled expansion state (backwards compatible)
- ProductionsTable and LocationsTable lift expanded state and render a 'Collapse all' button when rows are expanded
- Added translations for en and nl

image

### DIFF
--- a/frontend/next.config.ts
+++ b/frontend/next.config.ts
@@ -67,6 +67,7 @@ const nextConfig: NextConfig = {
     },
 
     // Production build
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     webpack(config: Configuration, options) {
         // Ensure module and rules exist on the config object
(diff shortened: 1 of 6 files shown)


frontend/next.config.ts
```diff
@@ -67,6 +67,7 @@ const nextConfig: NextConfig = {
     },
 
     // Production build
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Thoosje: Just remove the options, not used anymore if you want. But both are fine.

Thoosje (approved):

Kaya-Sem (approved):

PR #361, opened 2026-04-21
feat(cms): improved style and UX of collection picker dialog

Kaya-Sem:
Reworks the collection picker dialog's style and UX.

- Search-to-create: the search input now doubles as a new collection name: typing a query with no exact match shows a "+ Maak 'X' aan" row at the top of the list; clicking it creates and selects the collection. The separate create input/button section is removed.
- Buttons restyled to match the action bar (monospaced, uppercase, square, `sm` size); "Add and open" button removed.
- Inputs and list container are square (no border-radius); focus ring removed from inputs.
- List box has a fixed height (`min-h-56 max-h-56`) to prevent layout shift while filtering.
- `DialogContent` gets `grid-cols-1` to prevent content overflow on narrow viewports.
- Open/close animations removed from `DialogContent` and `DialogOverlay` to fix a pixel-shift glitch on close.
- `cursor-pointer` added to close button and collection list rows.

---

image

image

### DIFF
--- a/frontend/src/components/cms/collection-picker-dialog.tsx
+++ b/frontend/src/components/cms/collection-picker-dialog.tsx
@@ -142,7 +142,7 @@ export function CollectionPickerDialog({
                         {showCreate && (
                             <button
                                 type="button"
-                                className="text-muted-foreground hover:bg-muted block w-full cursor-pointer px-2 py-1 text-left text-sm"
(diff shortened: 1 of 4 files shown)


Thoosje (approved):

tst-038 (changes requested):
Image on hover text aint readable (make collection button)

tst-038:
image
for existing collections the text should be light colored (see table header)

tst-038 (approved):

PR #363, opened 2026-04-22
fix: s3 upload

tst-038:
Fixes S3 upload by reverting to previous versions of AWS S3 package


tst-038:
closed in favor of #364

PR #373, opened 2026-04-25
feat(frontend): improved homepage

Kaya-Sem:
Adds a published articles section to the homepage and tightens the featured productions grid.

- New `ArticlesSection` component shows up to 3 published articles in the same column-rule grid style as the featured section, with cover image, publish date, and title
- Featured card hierarchy fixed: artist was rendered as a faded duplicate of the title at the same size - now an italic byline; tagline gets a border-top and reads as a deck
- Primary featured card image taller (260px vs 200px) for more visual weight
- Removed the 4 genre tag pills under the search bar (they were text queries, not facet filters)

image


Thoosje (approved):

tst-038 (approved):
Image Since you're at it, this highlight is a bit to much and not consistent with other pages / components, might as well fix this

